PR #7, opened 2019-07-11
Fix build issues on Ubuntu 16.04

fabianoboril:
This PR fixes several build issues experienced on Ubuntu 16.04

* The required cmake version is downgraded from 3.7 to 3.5, as 3.7 seems not required and 3.5 is the default in Ubuntu 16.04 (which is supported according to the documentation)
* Linking of the OpenSceneGraph libraries did not work completely, so the linking flags were updated

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 message(STATUS "CMake version: " ${CMAKE_VERSION})
 
 # Set to the current lowest tested version of CMake
-cmake_minimum_required (VERSION 3.7.0 FATAL_ERROR)
+cmake_minimum_required (VERSION 3.5.0 FATAL_ERROR)
 
 project (EnvironmentSimulator)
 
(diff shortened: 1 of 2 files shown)


eknabevcc:
Regarding library linking: 
I now realize that the OpenSceneGraph (OSG) version has incremented to 3.7 for the Windows builds. But the instruction for Linux build says 3.6.3. 
My suggestion is to update Linux instruction to 3.7 as well, and keep that part of cmake in common. I don't know if anyone has tried to build OSG 3.7 for Linux, but I would expect it to work as good. Or do you see any reasons to stay on 3.6.3? In that case we just add an if-statement for Linux. But best is to stay in sync between Windows and Linux I think. If you agree upon switching to OSG 3.7, please update Linux instruction: https://github.com/esmini/esmini/blob/master/BuildOnLinux.md. Thanks!

Regarding cmake version, the reason for 3.7 was some issue with VisualStudio 2017 generator that seems to be fixed by 3.7. https://cmake.org/cmake/help/v3.7/release/3.7.html#generators
I propose we add an if-statement for this - it makes sense to support cmake default version for Linux. Something like this:
#if(UNIX)
  cmake_minimum_required (VERSION 3.5.0 FATAL_ERROR)
#else
  cmake_minimum_required (VERSION 3.7.0 FATAL_ERROR)
#endif()

PR #16, opened 2020-04-28
Fixed an issue where Position::Delta would ignore junctions

brifsttar:
Hi,

We've been using esmini internally for various projects, and we first must thank you for your amazing work.

We recently found an issue with `Position::Delta`, where the recursive calls to `Position::FindDistToPos` can sometimes wrongly return no match when the `RoadLink` is a junction. Indeed, line 3650, we have the following call:

`junction->GetNoConnectionsFromRoadId(road->GetId())`

Where `road` is initialized as the road we currently are on, not the road we were considering while exploring adjacents roads in `Position::Delta`. This means that that `GetNoConnectionsFromRoadId` can receive as a parameter a road that is *not* actually connecting into the junction, meaning we fail to check the junction for possible position match.

Our idea for the fix was to somehow get the actual incoming road. It would have been great if the `RoadLink` class had that information, but sadly it wasn't available. We therefore chose to add a parameter to the function, which doesn't have too much impact on existing code as only `Position::Delta` uses `Position::FindDistToPos`.

Also, since in a junction `RoadLink::GetContactPointType` can't be used to guess the contact point (it varies for each connecting road in the junction), we had to remove those calls and replace them with individual road contact point check.

We haven't thoroughly tested this fix, but it worked on our simple test case (which lead us to discover the issue in the first place).


eknabe:
Hi,
Thanks a lot for this highly valuable feedback. And not only reporting a problem but proposing solution as well. highly appreciated :). I have done some preliminary tests, on a new road model created for this purpose involving more than one intersection, during which I found additional issues. 
So I will try spend some time today sorting out both solution and how to deal with pull request correctly, including potentially adding commits (I'm kind of a GIT newbie).

eknabevcc:
Hi,
Thanks a lot for this highly valuable feedback. And not only reporting a problem but proposing solution as well. highly appreciated :). I have done some preliminary tests, on a new road model created for this purpose involving more than one intersection, during which I found additional issues.
So I will try spend some time today sorting out both solution and how to deal with pull request correctly, including potentially adding commits (I'm kind of a GIT newbie).

eknabevcc:
Indeed the path finding implementation had a lot of shortcomings. The proposed fixes is a good step forward. However when I tested on a bit more complex road network including multiple junctions I realized more issues (not in your proposal but in the original code). 
Based on your input I have now implemented a more generic (shortest) path finding, based on [dijkstra's algorithm](https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm). The implementation is, so far, in a separate branch [path_find_method](https://github.com/esmini/esmini/tree/find_path_method). 
I also added a new road model (multi_intersections.xodr) and added a corresponding 3D model (multi_intersections.osgb) to the [model package](https://drive.google.com/uc?export=download&id=11a8njhkRIjLYMqCsSL9pU-d5_A8pMVhc).
I realized that your change is done in a pretty old branch. And I'm not sure how to best go forward. I would appreciate if you could test the new branch. But since it is based on latest master I suspect it has diverged somewhat from your code and hence might involve some work on your side. Another option would be for us to continue on your branch. It would be nice though to pull you up to current esmini status.
What do you think is our best next step ahead?

brifsttar:
Wow, thanks for that patch, great work!

We have an esmini fork in a private repo, which I really hope we'll make public at some point. I try to keep it up-to-date with esmini's releases, so I just did that to merge your new branch. I'll get a colleague to test that on monday, since she was the one having the issue in the first place. I'll keep you updated, but I know that her test case is fairly simple, so it won't provide that much coverage.

brifsttar:
Hi,

My colleague and I tested your branch, it works perfectly fine on our (small) test case. I'll let you know if we have more test scenarios in the future.

eknabe:
Great, thanks for feedback!

Issue #17, opened 2020-05-01
ParamPoly3::EvaluateDS() returns wrong value on a specific geom?


brifsttar:
Hi,

I have an *.xodr which have a geometry that somehow makes ParamPoly3::EvaluateDS() go crazy. Since I can't share the whole *.xodr, here's the geometry line that causes issues:

```
<geometry s="0" x="-1096.3684428373817" y="-533.4096859078702" hdg="3.6849934514250173" length="10.836506271255859">
    <paramPoly3 pRange="normalized" aU="0" bU="12.024944062662236" cU="-9.0567143281122586" dU="2.1188575530043776" aV="0" bV="-1.7763568394002505e-15" cV="-15.827183832673695" dV="7.4526055111209653" />
</geometry>
```

The road along the geometry in RoadRunner looks like this.

![RoadRunner](https://i.imgur.com/HHwiPFB.png)

I used Position::MoveAlongS() to travel along this road (bottom lane, no offset, 1m step), and plotted the resulting X/Y coordinates

![Plot](https://i.imgur.com/Z8tSdGq.png)

Obviously, that's not what I expected. Both "straight" extremities are predecessor/successor, I kept them in this plot as reference. I looked at track coordinates (S, T) and they looked fine. Only X, Y and heading seemed wrong.

I tried to debug it myself, looking around ParamPoly3::EvaluateDS() to see what could be wrong, but my poor math skills failed me.

brifsttar:
I plotted more things, and I managed to reproduce the issue in RoadRunner if I remove those lines from the road declaration

```xml
<laneOffset s="0" a="8.0470581516279562" b="-0.0015518285952879777" c="1.3711908013277687" d="-1.5343170713315633" />
<laneOffset s="0.88146497931321333" a="8.0602540195476475" b="-1.1606418665715443" c="0.54585854125706335" d="-0.039166357212443305" />
<laneOffset s="7.525141535954571" a="12.957466728876136" b="0.90614781025203817" c="-0.45698342941205722" d="0.064498391536051553" />
```

So I guess somewhere in RoadManager the lane offset is forgotten? I'm heading back to debugging!

brifsttar:
More info on my fix in #18.

The heading returned by ParamPoly3::EvaluateDS() is causing the issues downstream. The XY coordinates look all right, but the heading is way off. Here is the returned heading, plotted for each meter of the road (angle in degree).

![Plot](https://i.imgur.com/8jlzAUy.png)

[...]

PR #18, opened 2020-05-02
[roadmngr] Fixed wrong heading being computed returned by ParamPoly3:…

brifsttar:
My fix for #17. There might be a more elegant way to compute heading, but I'm bad at math :-/

### DIFF
--- a/EnvironmentSimulator/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/RoadManager/RoadManager.cpp
@@ -255,7 +255,7 @@ void ParamPoly3::EvaluateDS(double ds, double *x, double *y, double *h)
 
 	*x = GetX() + u_local * cos(GetHdg()) - v_local * sin(GetHdg());
 	*y = GetY() + u_local * sin(GetHdg()) + v_local * cos(GetHdg());
-	*h = GetHdg() + atan2(v_local, u_local);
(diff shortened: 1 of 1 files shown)


eknabevcc:
Thanks a lot for reporting this bug. I've seen it before but hadn't opportunity to investigate yet. 
The root cause of the bug is that in all roads we've used so far included only the arcLength parameter range of ParamPoly3. See http://www.opendrive.org/docs/OpenDRIVEFormatSpecRev1.4H.pdf page 46. And hence the Normalized version (actually default) has not been tested. 
What is very strange though is that omitting the atan (or atan2) operation still produced OK road lanes (at least they looked OK)... I can't understand how arcLength vs Normalized leads to such different behavior. Well, I'm not a mathematician either and it doesn't matter anyway. 
The poly3V_.EvaluatePrim(ds) is evaluating the derivative of the V-dimension (local y). So the expression poly3V_.EvaluatePrim(ds)/poly3U_.EvaluatePrim(ds) is like dY/dX. Which basically is what you proposed with local_V/local_U, but potentially more stable at edge cases.
The bug was that the slope was calculated but NOT the heading, which was a big mistake. The atan operation was the missing link. So thanks for finding it and pointing out the solution!

eknabevcc:
Just realized there is additional adjustments needed, i.e. scaling of the ds/p parameter. It will be committed to the master later today.

PR #33, opened 2020-08-21
Add data to OSI orientationRate.yaw

Eidland:
By using previous heading and the new heading
the orientation rate is calculated.
The orientation rate is set after the MoveAlongS function
which steps each object forward so we are sure
we are getting the new heading.
The heading rate is found in the position object along
with similar data, such as heading and velocity.


eknabevcc:
Approved. Merge pull request.

PR #36, opened 2020-08-25
Add csv output

CPettit-LCM:
First pull request for adding a CSV logging function to ESMINI

The Linz Center of Mechatronics is incorporating ESMINI into a workflow automation tool called [SyMSpace ](https://symspace.lcm.at/) and required an ASCII output for post processing of simulated OpenScenario tests.

The launch argument handling was added to PlayerBase.

EgoSimulator main now has an UpdateCSV_Log method that is called every timestep. It only executes if CSV_Log is not null. It could be moved into UpdateExternVehicles. 

The main body of the CSV logger is implemented side by side with the main Logger class in CommonMini.


eknabevcc:
Excellent. Easy to merge. I made a just few minor adjustments.
Thanks for contribution!

PR #38, opened 2020-09-04
Fixed bugs in Position::GetDistToTrackGeom

brifsttar:
Hi Emil,

I've encountered a few bugs in Position::GetDistToTrackGeom, so here's a PR. First I'll explain the changes, then I'll go over the example that lead me to those bugs.

The first fix is similar to #19, where the lane offset is wrongly computed from road start instead of geometry start.

The second fix is a bit more complicated, and I'm not entierly satisfied with it, nor have  I thoroughly tested it. The thing is that for points outside the geom, the "exact distance" is still computed, even though there is a comment saying it is not (https://github.com/brifsttar/esmini/blob/e453b9b96980fbb5442399a5e1d3336f5accef02/EnvironmentSimulator/RoadManager/RoadManager.cpp#L4490). This leads to all kinds of weird things happening (more on that in the example).

The third fix is that the method only tries to compute distances on "driving" lanes, which seems a bit arbitrary (and definitely causes me issues, as I'm working with sidewalks).

So my problem was that I tried to map a point to track coordinate using roadmanager::Position. The point is on the sidewalk of the road first geom. But due to the road configuration, the sidewalk is pretty far from the closest driving lane. Since the method only tried to match on driving lanes, the minimal distance found on that geometry was pretty high (~7m).

Then it evaluated the second geometry. It correctly found the point not ```inside```, at a ~6m distance. But that distance is not the lateral (i.e. T) distance, but the distance between the reference point, and the start of the geometry. Which, as it happens in my example, is mostly longitudinal (i.e. S). However, the rest of the algorithm expects ```dist``` to be the lateral distance.

So then, even though the point is outside and dist is NOT lateral distance, it tries to match that ```dist``` onto lanes. And in my case, I have a (driving) lane at a lateral offset of ~7m, so the algorithm sets the ```min_lane_dist``` at 7 (lane offset) - 6 (point offset) = ~1m within the driving lane. When the point actually is on a sidewalk, ~6m behind (longitudinal).

This leads to this value (~1m) being returned to ```XYZH2TrackPos```, where even with the added weight due to being outside (2), this ```dist``` is still lower than the one from the previous (correct) geometry, since this one was computed with driving lane only (and we're on a very wide sidewalk).

I'm pretty confident that I lost you in my explanation, so I also used all of my mspaint skills to draw something (position of geom/points is not accurate).

The yellow line seperates the two geoms. The green dot is the point I try to get track coordinate. The blue line is the ```dist``` used by the method to match on the first geom, and the orange line is for the second geom. Since the orange is shorter than the blue, it wrongly positions me on the second geom, which leads to all kind of issues.

Below is the coordinate for the green point, and the road description.

If you want a live explanation of all that, it's possible :-)

```
x : 188,20 y : 96,900 z : 108,0
```

```xml
    <road name="Road 161" length="4.3844816182604788e+1" id="161" junction="-1">
        <link>
            <predecessor elementType="junction" elementId="2049"/>
            <successor elementType="junction" elementId="189"/>
        </link>
        <type s="0.0000000000000000e+0" type="town">
            <speed max="50" unit="km/h"/>
        </type>
        <planView>
            <geometry s="0.0000000000000000e+0" x="-2.0418483954313496e+2" y="9.4849797419553639e+1" hdg="3.7632493637335163e-1" length="2.6758439213118052e+1">
                <line/>
            </geometry>
            <geometry s="2.6758439213118052e+1" x="-1.7929891529954256e+2" y="1.0468365951256234e+2" hdg="3.7632493637335163e-1" length="2.3201874848681747e+0">
                <arc curvature="2.0000000000000000e-3"/>
            </geometry>
            <geometry s="2.9078626697986223e+1" x="-1.7714307713347304e+2" y="1.0554134366584879e+2" hdg="3.8096531134313882e-1" length="1.4766189484618565e+1">
                <line/>
            </geometry>
        </planView>
        <elevationProfile>
            <elevation s="0.0000000000000000e+0" a="0.0000000000000000e+0" b="0.0000000000000000e+0" c="0.0000000000000000e+0" d="0.0000000000000000e+0"/>
            <elevation s="3.7401783502158139e+0" a="0.0000000000000000e+0" b="0.0000000000000000e+0" c="0.0000000000000000e+0" d="0.0000000000000000e+0"/>
            <elevation s="3.3740178350215814e+1" a="0.0000000000000000e+0" b="0.0000000000000000e+0" c="0.0000000000000000e+0" d="0.0000000000000000e+0"/>
        </elevationProfile>
        <lateralProfile>
            <superelevation s="0.0000000000000000e+0" a="0.0000000000000000e+0" b="0.0000000000000000e+0" c="0.0000000000000000e+0" d="0.0000000000000000e+0"/>
        </lateralProfile>
        <lanes>
            <laneOffset s="0.0000000000000000e+0" a="-5.0350000000000001e+0" b="1.5067848297979789e-18" c="-1.1178614589718686e-19" d="1.7657692776074944e-21"/>
            <laneSection s="0.0000000000000000e+0" singleSide="false">
                <left>
                    <lane id="7" type="sidewalk" level="false">
                        <width sOffset="0.0000000000000000e+0" a="2.0000000000000000e+0" b="0.0000000000000000e+0" c="0.0000000000000000e+0" d="0.0000000000000000e+0"/>
                        <roadMark sOffset="0.0000000000000000e+0" type="none" material="standard" color="white" laneChange="none"/>
                        <speed sOffset="0.0000000000000000e+0" max="5.0000000000000000e+1" unit="km/h"/>
                        <userData>
                            <vectorLane sOffset="0.0000000000000000e+0" laneId="{26a21073-b2e2-4385-9497-9a7701956839}" travelDir="undirected"/>
                        </userData>
                    </lane>
                    <lane id="6" type="shoulder" level="false">
                        <width sOffset="0.0000000000000000e+0" a="6.3499999999999979e-1" b="-1.3456990696216876e-18" c="-6.6080917439023062e-20" d="8.8996486619208633e-22"/>
                        <roadMark sOffset="0.0000000000000000e+0" type="curb" material="standard" width="1.5239999999999998e-1" laneChange="none"/>
                        <speed sOffset="0.0000000000000000e+0" max="5.0000000000000000e+1" unit="km/h"/>
                        <userData>
                            <vectorLane sOffset="0.0000000000000000e+0" laneId="{0899fc78-7e7c-4d96-9247-1f4fd558d920}" travelDir="undirected"/>
                        </userData>
                    </lane>
                    <lane id="5" type="shoulder" level="false">
                        <width sOffset="0.0000000000000000e+0" a="5.0000000000000000e-1" b="0.0000000000000000e+0" c="0.0000000000000000e+0" d="0.0000000000000000e+0"/>
                        <roadMark sOffset="0.0000000000000000e+0" type="none" material="standard" color="white" laneChange="none"/>
                        <speed sOffset="0.0000000000000000e+0" max="5.0000000000000000e+1" unit="km/h"/>
                        <userData>
                            <vectorLane sOffset="0.0000000000000000e+0" laneId="{f22d9c9f-4fbd-4cca-84cb-2f5a66d57945}" travelDir="undirected"/>
                        </userData>
                    </lane>
                    <lane id="4" type="driving" level="false">
                        <width sOffset="0.0000000000000000e+0" a="3.5000000000000000e+0" b="0.0000000000000000e+0" c="0.0000000000000000e+0" d="0.0000000000000000e+0"/>
                        <roadMark sOffset="0.0000000000000000e+0" type="none" material="standard" color="white" laneChange="none"/>
                        <speed sOffset="0.0000000000000000e+0" max="5.0000000000000000e+1" unit="km/h"/>
                        <userData>
                            <vectorLane sOffset="0.0000000000000000e+0" laneId="{9613e0d4-181a-4261-ad14-2c7d5b5e8e9b}" travelDir="backward"/>
                        </userData>
                    </lane>
                    <lane id="3" type="parking" level="false">
                        <width sOffset="0.0000000000000000e+0" a="2.3999999999999999e+0" b="0.0000000000000000e+0" c="0.0000000000000000e+0" d="0.0000000000000000e+0"/>
                        <roadMark sOffset="0.0000000000000000e+0" type="none" material="standard" color="white" laneChange="none"/>
                        <speed sOffset="0.0000000000000000e+0" max="5.0000000000000000e+1" unit="km/h"/>
                        <userData>
                            <vectorLane sOffset="0.0000000000000000e+0" laneId="{d340caf0-b8db-4265-8f80-7f53a3f79ecc}" travelDir="undirected"/>
                        </userData>
                    </lane>
                    <lane id="2" type="shoulder" level="false">
                        <width sOffset="0.0000000000000000e+0" a="6.3500000000000023e-1" b="-1.5067848297979789e-18" c="-5.7687296792709616e-20" d="7.8232106215609829e-22"/>
                        <roadMark sOffset="0.0000000000000000e+0" type="none" material="standard" color="white" laneChange="none"/>
                        <speed sOffset="0.0000000000000000e+0" max="5.0000000000000000e+1" unit="km/h"/>
                        <userData>
                            <vectorLane sOffset="0.0000000000000000e+0" laneId="{8798b09e-ac77-4c03-ade3-28b8856c391f}" travelDir="undirected"/>
                        </userData>
                    </lane>
                    <lane id="1" type="sidewalk" level="false">
                        <width sOffset="0.0000000000000000e+0" a="2.0000000000000000e+0" b="0.0000000000000000e+0" c="0.0000000000000000e+0" d="0.0000000000000000e+0"/>
                        <roadMark sOffset="0.0000000000000000e+0" type="curb" material="standard" width="1.5239999999999998e-1" laneChange="none"/>
                        <speed sOffset="0.0000000000000000e+0" max="5.0000000000000000e+1" unit="km/h"/>
                        <userData>
                            <vectorLane sOffset="0.0000000000000000e+0" laneId="{48ab06f5-95a6-4619-87c7-62ee546b886c}" travelDir="undirected"/>
                        </userData>
                    </lane>
                </left>
                <center>
                    <lane id="0" type="none" level="false">
                        <roadMark sOffset="0.0000000000000000e+0" type="none" material="standard" color="white" laneChange="none"/>
                        <userData/>
                    </lane>
                </center>
                <userData>
                    <vectorLaneSection>
                        <carriageway rightBoundary="0" leftBoundary="0"/>
                        <carriageway rightBoundary="2" leftBoundary="6"/>
                    </vectorLaneSection>
                </userData>
            </laneSection>
        </lanes>
        <objects>
            <object id="3724" name="" s="2.2370522818614873e+0" t="-1.1999960287965052e+0" zOffset="0.0000000000000000e+0" hdg="-1.5707978010177612e+0" roll="0.0000000000000000e+0" pitch="0.0000000000000000e+0" orientation="+" type="parking" width="5.6137487388483009e+0" length="2.4000497443380482e+0">
                <parkingSpace>
                    <marking side="left" type="solid" width="1.2500000000000000e-1" color="white"/>
                    <marking side="right" type="solid" width="1.2500000000000000e-1" color="white"/>
                </parkingSpace>
            </object>
            <object id="3678" name="" s="7.8508094988603379e+0" t="-1.1999960287965337e+0" zOffset="0.0000000000000000e+0" hdg="-1.5707914829254150e+0" roll="0.0000000000000000e+0" pitch="0.0000000000000000e+0" orientation="-" type="parking" width="5.6137811005903586e+0" length="2.4000777932870392e+0">
                <parkingSpace>
                    <marking side="left" type="solid" width="1.2500000000000000e-1" color="white"/>
                    <marking side="right" type="solid" width="1.2500000000000000e-1" color="white"/>
                </parkingSpace>
            </object>
            <object id="3680" name="" s="1.3464566715859196e+1" t="-1.2000031242975524e+0" zOffset="0.0000000000000000e+0" hdg="-1.5707851648330688e+0" roll="0.0000000000000000e+0" pitch="0.0000000000000000e+0" orientation="-" type="parking" width="5.6137619562199461e+0" length="2.4000343150162280e+0">
                <parkingSpace>
                    <marking side="left" type="solid" width="1.2500000000000000e-1" color="white"/>
                    <marking side="right" type="solid" width="1.2500000000000000e-1" color="white"/>
                </parkingSpace>
            </object>
            <object id="3661" name="" s="1.9078295550853763e+1" t="-1.1999960287965052e+0" zOffset="0.0000000000000000e+0" hdg="-1.5708023309707642e+0" roll="0.0000000000000000e+0" pitch="0.0000000000000000e+0" orientation="+" type="parking" width="5.6137369875198999e+0" length="2.4000181614143230e+0">
                <parkingSpace>
                    <marking side="left" type="solid" width="1.2500000000000000e-1" color="white"/>
                    <marking side="right" type="solid" width="1.2500000000000000e-1" color="white"/>
                </parkingSpace>
            </object>
            <object id="3664" name="" s="2.4694206247559720e+1" t="-1.1993188671130781e+0" zOffset="0.0000000000000000e+0" hdg="-1.5689975023269653e+0" roll="0.0000000000000000e+0" pitch="0.0000000000000000e+0" orientation="-" type="parking" width="5.6181360868327488e+0" length="2.4084007378830279e+0">
                <parkingSpace>
                    <marking side="left" type="solid" width="1.2500000000000000e-1" color="white"/>
                    <marking side="right" type="solid" width="1.2500000000000000e-1" color="white"/>
                </parkingSpace>
            </object>
            <object id="3229" name="arrow" s="2.4890178350215781e+1" t="1.7500014136988113e+0" zOffset="0.0000000000000000e+0" hdg="3.1415927410125732e+0" roll="0.0000000000000000e+0" pitch="0.0000000000000000e+0" orientation="+" type="-1" width="1.5739274955203655e+0" length="1.4896625317754797e+0"/>
            <object id="3703" name="" s="3.0305782420943402e+1" t="-1.1984740151482356e+0" zOffset="0.0000000000000000e+0" hdg="-1.5708150863647461e+0" roll="0.0000000000000000e+0" pitch="0.0000000000000000e+0" orientation="+" type="parking" width="5.6205357094759165e+0" length="2.4035673571115694e+0">
                <parkingSpace>
                    <marking side="left" type="solid" width="1.2500000000000000e-1" color="white"/>
                    <marking side="right" type="solid" width="1.2500000000000000e-1" color="white"/>
                </parkingSpace>
            </object>
            <object id="3698" name="" s="3.5919538600040632e+1" t="-1.2000012626155865e+0" zOffset="0.0000000000000000e+0" hdg="-1.5708070993423462e+0" roll="0.0000000000000000e+0" pitch="0.0000000000000000e+0" orientation="+" type="parking" width="5.6137832916110995e+0" length="2.4000308445096152e+0">
                <parkingSpace>
                    <marking side="left" type="solid" width="1.2500000000000000e-1" color="white"/>
                    <marking side="right" type="solid" width="1.2500000000000000e-1" color="white"/>
                </parkingSpace>
            </object>
            <object id="3651" name="" s="4.1533315315094718e+1" t="-1.2000012626155865e+0" zOffset="0.0000000000000000e+0" hdg="-1.5707879066467285e+0" roll="0.0000000000000000e+0" pitch="0.0000000000000000e+0" orientation="-" type="parking" width="5.6137710596839412e+0" length="2.4000252235221069e+0">
                <parkingSpace>
                    <marking side="left" type="solid" width="1.2500000000000000e-1" color="white"/>
                    <marking side="right" type="solid" width="1.2500000000000000e-1" color="white"/>
                </parkingSpace>
            </object>
        </objects>
    </road>
```

### DIFF
--- a/EnvironmentSimulator/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/RoadManager/RoadManager.cpp
@@ -679,7 +679,7 @@ int Road::GetLaneSectionIdxByS(double s, int start_at)
 	return (int)i;
 }
 
-LaneInfo Road::GetLaneInfoByS(double s, int start_lane_section_idx, int start_lane_id)
+LaneInfo Road::GetLaneInfoByS(double s, int start_lane_section_idx, int start_lane_id, Lane::LaneType laneType)
 {
 	LaneInfo lane_info;
 	
@@ -716,13 +716,13 @@ LaneInfo Road::GetLaneInfoByS(double s, int start_lane_section_idx, int start_la
 				}
 			}
 
-			// If new lane is not driving, try to move into a close driving lane
-			if (!lane_section->GetLaneById(lane_info.lane_id_)->IsDriving())
+			// If new lane is not of snapping type, try to move into a close valid lane
+			if (laneType & lane_section->GetLaneById(lane_info.lane_id_)->GetLaneType())
 			{
-				lane_info.lane_id_ = lane_section->FindClosestDrivingLane(lane_info.lane_id_);
+				lane_info.lane_id_ = lane_section->FindClosestSnappingLane(lane_info.lane_id_, laneType);
 				if (lane_info.lane_id_ == 0)
 				{
-					LOG("Failed to find a closest driving lane");
+					LOG("Failed to find a closest snapping lane");
 				}
 			}
 		}
@@ -808,14 +808,14 @@ Lane* LaneSection::GetLaneById(int id)
 	return 0;
 }
 
-int LaneSection::FindClosestDrivingLane(int id)
+int LaneSection::FindClosestSnappingLane(int id, Lane::LaneType laneType)
 {
 	int id_best = id;
 	size_t delta_best = lane_.size() + 1;
 
 	for (size_t i = 0; i < lane_.size(); i++)
 	{
-		if (lane_[i]->IsDriving())
+		if (laneType & lane_[i]->GetLaneType())
 		{
 			if ((abs(lane_[i]->GetId() - id) < delta_best) || 
 				(abs(lane_[i]->GetId() - id) == delta_best && abs(lane_[i]->GetId()) < abs(id_best)))
@@ -5436,7 +5436,7 @@ int Position::SetLanePos(int track_id, int lane_id, double s, double offset, int
 	}
 	else  // Find LaneSection and info according to s
 	{
-		LaneInfo lane_info = road->GetLaneInfoByS(s_, lane_section_idx_, lane_id_);
+		LaneInfo lane_info = road->GetLaneInfoByS(s_, lane_section_idx_, lane_id_, snapToLaneTypes_);
 		lane_section_idx_ = lane_info.lane_section_idx_;
 		lane_id_ = lane_info.lane_id_;
 		
@@ -5515,7 +5515,7 @@ void Position::SetLaneBoundaryPos(int track_id, int lane_id, double s, double of
 	}
 	else  // Find LaneSection and info according to s
 	{
-		LaneInfo lane_info = road->GetLaneInfoByS(s_, lane_section_idx_, lane_id_);
+		LaneInfo lane_info = road->GetLaneInfoByS(s_, lane_section_idx_, lane_id_, snapToLaneTypes_);
 		lane_section_idx_ = lane_info.lane_section_idx_;
 		lane_id_ = lane_info.lane_id_;
 		
@@ -5604,7 +5604,7 @@ void Position::SetRoadMarkPos(int track_id, int lane_id, int roadmark_idx, int r
 	}
 	else  // Find LaneSection and info according to s
 	{
-		LaneInfo lane_info = road->GetLaneInfoByS(s_, lane_section_idx_, lane_id_);
+		LaneInfo lane_info = road->GetLaneInfoByS(s_, lane_section_idx_, lane_id_, snapToLaneTypes_);
 		lane_section_idx_ = lane_info.lane_section_idx_;
 		lane_id_ = lane_info.lane_id_;
 		
--- a/EnvironmentSimulator/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/RoadManager/RoadManager.cpp
@@ -5360,11 +5360,10 @@ int Position::MoveAlongS(double ds, double dLaneOffset, Junction::JunctionStrate
 	// the SIGN() adjustment. But for now this adjustment means that a positive dLaneOffset always moves left?
 	offset_ += dLaneOffset * -SIGN(GetLaneId());
 	double s_stop = 0;
+	ds_signed = -SIGN(GetLaneId()) * ds; // adjust sign of ds according to lane direction - right lane is < 0 in road dir
 	
 	for (int i = 0; i < max_links; i++)
 	{
-		ds_signed = -SIGN(GetLaneId()) * ds; // adjust sign of ds according to lane direction - right lane is < 0 in road dir
-
 		if (s_ + ds_signed > GetOpenDrive()->GetRoadByIdx(track_idx_)->GetLength())
 		{
 			ds_signed = s_ + ds_signed - GetOpenDrive()->GetRoadByIdx(track_idx_)->GetLength();
@@ -5390,7 +5389,12 @@ int Position::MoveAlongS(double ds, double dLaneOffset, Junction::JunctionStrate
 			return ErrorCode::ERROR_END_OF_ROAD;
 		}
 
-		ds = SIGN(ds) * fabs(ds_signed);
+		// Roads aren't going the same direction, we flip ds
+		if ((link->GetType() == LinkType::PREDECESSOR && contact_point_type == ContactPointType::CONTACT_POINT_START) ||
+		    (link->GetType() == LinkType::SUCCESSOR && contact_point_type == ContactPointType::CONTACT_POINT_END))
+		{
+			ds_signed *= -1;
+		}
 	}
 
 	SetLanePos(track_id_, lane_id_, s_ + ds_signed, offset_);
--- a/EnvironmentSimulator/RoadManager/RoadManager.hpp
+++ b/EnvironmentSimulator/RoadManager/RoadManager.hpp
@@ -590,7 +590,7 @@ namespace roadmanager
 		double GetS() { return s_; }
 		Lane* GetLaneByIdx(int idx);
 		Lane* GetLaneById(int id);
-		int FindClosestDrivingLane(int id);
+		int FindClosestSnappingLane(int id, Lane::LaneType laneType = Lane::LaneType::LANE_TYPE_ANY_DRIVING);
 		int GetLaneIdByIdx(int idx);
 		int GetLaneIdxById(int id);
 		int GetLaneGlobalIdByIdx(int idx);
@@ -804,7 +804,7 @@ namespace roadmanager
 		*/
 		double GetCenterOffset(double s, int lane_id);
 
-		LaneInfo GetLaneInfoByS(double s, int start_lane_link_idx, int start_lane_id);
+		LaneInfo GetLaneInfoByS(double s, int start_lane_link_idx, int start_lane_id, Lane::LaneType laneType = Lane::LaneType::LANE_TYPE_ANY_DRIVING);
 		double GetLaneWidthByS(double s, int lane_id);
 		double GetSpeedByS(double s);
 		bool GetZAndPitchByS(double s, double *z, double *pitch, int *index);


eknabevcc:
Wow! Thanks a lot for this excellent issue report including detailed explanation, figure and even scenario to reproduce it. 
These distance calculations are tricky and I'm very happy if we can find and solve any problems within.
I realize that reading this through Friday evening is not enough :) I will look closer and also review whats actually implemented and then get back.

eknabevcc:
I'm looking into this now. The third commit, fixing bad lane offset, is spot on a bug fix. Great finding!

The second one, matching also non driving lanes, is a bit tricky. The idea has been to find the closest "valid" lane for a vehicle. But so far esmini has not dealt with pedestrians and bicycles (soon to be though). And I guess there can be other use cases where you want to match non driving lanes as well. May I ask what use case you have for this? 
I'm thinking that perhaps it could be an option to the function (XYZH2TrackPos) whether to consider all lanes or only driving, or even specifying what lane types to match for?
So I'm a little bit cautious about this fix.

Then the first one. As you say, this is tricky. And I admit it's poorly documented in the code or elsewhere. The point may be "inside" the segment when approximated by a straight line. The second "inside" calculation is whether it is inside the "fan" shape defined by the normals at segment endpoints. See image below:
image
So I think that points outside the "fan" does not automatically disqualifies from the final calculations.

Anyway, when looking closer to this topic and testing some "challenging" OpenDRIVE files, I found:

1. A few additional bugs.
2. The method does not work very well for drastic clothoids and curves. 

So actually I started to think about a different method based on subsampling the road into linear segments. I will do some experiments before deciding way forward.

Finally I really want to thank you for this issue report, since it brings into focus a weak part in esmini that I want to improve. 
Please think and get back regarding lane type while I experiment with method.

brifsttar:
> May I ask what use case you have for this?

We're making a very basic pedestrian controller to populate our scene (within Unreal Engine). The idea is to have pedestrians automatically walk along sidewalks. To do that we place a "target point" on the sidewalk, and then move it forward with MoveAlongS as the pedestrian walks closer to it (using UE's AI system).

> I'm thinking that perhaps it could be an option to the function (XYZH2TrackPos) whether to consider all lanes or only driving, or even specifying what lane types to match for?

That would work, but this change doesn't need to be merged. We're working with our own fork of esmini, so we already have some changes on our side that are totally custom for us, this one can be part of it. I put it in the PR to ask whether the current limitation (i.e. driving lane only) was voluntary, or if it was more of a leftover from an early prototype that just stayed here. As long as you're aware of the limitation, I'm fine with whatever you decide.

> So actually I started to think about a different method based on subsampling the road into linear segments.

A few years ago I actually implemented a library that was somewhat similar to roadmanager (except my implementation was way, way less complete than yours), and I used subsampling for the geom projection. IIRC, I used dichotomy to first split the geom in two segments, find the best one, then split that half of the geom in two, etc. I admit I didn't really take the time to understand how GetDistToTrackGeom did that, which is why I simply added a ```return``` to fix my issue.

eknabevcc (approved):
Clearly a bug fix. Great finding, thanks!

eknabevcc:
cherry-picked [commit 3](https://github.com/esmini/esmini/commit/82312893f6d78009f2843d954a869667bfe21090).
closing pull request.

eknabevcc:
By the way. I replaced the old, shaky method to map world (x,y,z) positions to road network. The XYZH2TrackPos function has been rewritten from scratch with inspiration from your input; both the use case of snapping to custom selection of lanes and also using point sampling of the road geometry curves (actually re-using OSI-points). 
The lane type selection works as a bit-mask. So user can define any combination of lane types to "snapped" to.
I have tried it out on your OpenDRIVE example, and it seems OK. But it would be of high value if you could try out as well, to see if it works better for your use case.  
If you're using EgoSimulator, please check line 203 in ScenarioGateway.cpp. You can change the line from:
> obj_state->state_.pos.SetSnapLaneTypes(roadmanager::Lane::LaneType::LANE_TYPE_ANY_DRIVING);

to:

> obj_state->state_.pos.SetSnapLaneTypes(roadmanager::Lane::LaneType::LANE_TYPE_ANY);

But you can set it with any lane types, e.g:
> obj_state->state_.pos.SetSnapLaneTypes(roadmanager::Lane::LaneType::LANE_TYPE_BIKING | roadmanager::Lane::LaneType::LANE_TYPE_SIDEWALK);

brifsttar:
Hi Emil,

> If you're using EgoSimulator

As a side note, just to better explain what we're doing: we're only using roadmanager, on top of which we're developing a driver model. We interfaced both (roadmanager and our model) into our old simulation stack (custom), and to our new stack which runs on Unreal Engine. So I'm only familiar with the code in roadmanager.

This also explains why your new branch initially crashed our simulation: I was missing the call to OpenDrive::SetRoadOSI(), which I never needed until now. With no OSI points initialized, it was difficult to match any point onto the network! But after this was fixed, the first early test I ran (our demo scenario) was working as intenteded, so that's great news.

My colleague which initially stumbled onto the issues mentionned in this PR is currently on holidays, but I'll ask her to test your branch more thoroughly with all her scenarios that have pedestrians. I'll keep you updated.

Thank you for your amazing work on esmini and your wonderful support!

eknabevcc:
> This also explains why your new branch initially crashed our simulation: I was missing the call to OpenDrive::SetRoadOSI(), which I never needed until now. With no OSI points initialized, it was difficult to match any point onto the network! 

Good point! The code is now changed so that OSI points are always generated when loading an OpenDRIVE file. So it should not be necessary to explicitly call SetRoadOSI.

eknabevcc:
Just curious, did it work for your colleague as well? In that case I'll close this issue for now.

brifsttar:
She just came back today and ran the tests, it seems that her pedestrians have a lot more issues now. Some may be related to the changes induced by your fix, some may also be unrelated issues that become apparent now that the first issue was fixed. I'll investigate this tomorrow or friday, and I'll keep you updated.

eknabevcc:
Ok I see. Well, there is no hurry at all. Take your time. Any feedback is highly appreciated.

eknabevcc:
FYI: We just adjusted the XYZ2TrackPos (commit [8b5281b1](https://github.com/esmini/esmini/commit/8b5281b16f36f86b49f95ed2a7371c93d09566e9)) in a way that IF a vehicle is bound to a route the mapping will be constrained to that route. In effect the search for closest road coordinate will be limited to the road segments part of the route (actually searching all and only all waypoint road IDs).

Probably this does not affect your use case. But if it is the case that you assign routes to pedestrians this change could improve behavior.

brifsttar:
Hi Emil,

I fixed all the bugs that my colleague reported me. Only two of those were linked to esmini.

https://github.com/esmini/esmini/pull/38/commits/79e21ecc252c67ea41c581f0351216d81ed27ae7 is pretty straightforward. There was another part of the code where the lane type was assumed to be "driving", so I just replaced that with the new enum you added, it works perfectly.

https://github.com/esmini/esmini/pull/38/commits/1b22036e784d2fb5585b539f0fadd19867fd0625 is a bit trickier, and I had to go through several implementation before finding this fix. I make heavy use of MoveAlongS for my pedestrian controller, and I've had issues when MoveAlongS steps over a new road (let's say from road A to B, lane LA to LB). If A and B aren't in the same direction, but LA and LB are on the same side, MoveAlongS would keep jumping from A to B to A and so on.

Let's say you're on A, at ```s=0.5```, ```laneid=-1```. Desired ```ds=-1```. ```ds_signed``` is therefore -1. You jump straight into [```else if (s_ + ds_signed < 0)```](https://github.com/esmini/esmini/blob/master/EnvironmentSimulator/RoadManager/RoadManager.cpp#L5374), new ```ds_signed``` is -0.5.
At [```ds = SIGN(ds) * fabs(ds_signed)```](https://github.com/esmini/esmini/blob/master/EnvironmentSimulator/RoadManager/RoadManager.cpp#L5393), ```ds``` is updated to -0.5.
Next iteration, back at [```ds_signed = -SIGN(GetLaneId()) * ds```](https://github.com/esmini/esmini/blob/master/EnvironmentSimulator/RoadManager/RoadManager.cpp#L5366), ```ds_signed``` is updated to -0.5, since we're still on a negative laneid! And since ```s_``` is at 0 (A and B are connected "toe to toe"), we once again jump straight through [```else if (s_ + ds_signed < 0)```](https://github.com/esmini/esmini/blob/master/EnvironmentSimulator/RoadManager/RoadManager.cpp#L5374), and so on.

You might wonder in what weird use case would we have A and B connected, not in the same direction, but with LA and LB on the same side. Because that's definitely not a common use case! Well, the answer is: sidewalks.



The above image is taken from RoadRunner, with our test road network. I've highlighted the road geoms, and drawn arrows to show their direction. In this case, they're connected "head to head", but it's the same idea. The two dots on the sidewalks are both on a right-side lane.

So, after all that explaining, that still doesn't say anything about my fix. Well, I didn't fully understand what [```ds = SIGN(ds) * fabs(ds_signed)```](https://github.com/esmini/esmini/blob/master/EnvironmentSimulator/RoadManager/RoadManager.cpp#L5393) and [```ds_signed = -SIGN(GetLaneId()) * ds```](https://github.com/esmini/esmini/blob/master/EnvironmentSimulator/RoadManager/RoadManager.cpp#L5366) were supposed to achieve, so I removed them. I simply added a condition to check if roads are connected "head to head" or "toe to toe", in which case I change the sign on ```ds```. I'm definitely not 100% confident on that one. My colleague had initially spotted 13 pedestrians with issues in her scenario, and I spotted a few more trying to fix things. After applying this fix, I had no further issue. But all my use cases for MoveAlongS involve pedestrians and their weird junction connection, so I'd strongly advise to test it more with cars and more conventional lanes connections.

Also, not a bug, but something interesting to note: as I previously mentioned, we work within Unreal Engine, where single precision float is the norm. So when passing values from and to esmini, we have to cast from double to single to double precision, leading to some obvious rounding errors. This lead me to [this check](https://github.com/esmini/esmini/blob/master/EnvironmentSimulator/RoadManager/RoadManager.cpp#L5005), where I was passing an S-value received from a previous esmini call, but downcasted to single precision in-beetween. In such cases, ```SMALL_NUMBER``` isn't small enough to take into account those rounding errors, and the method would fail (which was problematic, as it's used by Position's constructor). I found a way to fix the issue within UE, but I thought it'd be good to share this, as it can possibly become an issue for you too later on.

eknabevcc:
Hi Richard!

Thanks for your excellent analysis and description of the problem. I think your use case with pedestrians is interesting. We're just starting similar use cases as well.

The first commit seems to be a clear fix. I must admit I didn't test the new lane type enum thoroughly, so I really appreciate your contribution!

Regarding the second commit, it's a bit trickier as you say :) Since it is a critical part of the roadmanager I have looked closer into it and also tried to reproduce it, however with no "success". I'll try to explain:
When two roads are connected head-to-head or toe-to-toe (thanks for analogy!) typically the sign of lane IDs changes. At least if they are on same side of reference line, which is not always the case - see pic below.
image

So, instead of relying on a general rule for the sign of the connected lane it is calculated. And that's the purpose of the line 

> ds_signed = -SIGN(GetLaneId()) * ds;

The idea is that ds should represent the desired distance to move along the road network. The sign indicates whether to go forward or backwards according to lane direction. So the other mystical calculation: 

> ds = SIGN(ds) * fabs(ds_signed);

has the purpose to update ds with the original sign but updated remaining distance to go. Note that the original ds might be huge so that the MoveAlongS function needs to cross multiple road connections in order to reach the distance.

Anyway, I really want to identify any bugs in this part of the roadmanager. I put together two minimalistic ODRs and OSCs to cover the two cases: head-to-head and toe-to-toe. And the car travels negative direction over the connection. Please see pic below and attached zip.
image


[two_roads.zip](https://github.com/esmini/esmini/files/5290664/two_roads.zip)

Conclusion: I'm definitely not sure, but I see a risk of the fix not covering all cases. 
I would be VERY interested in receiving your ODR. I would like to check it, and try to reproduce the problem. Please provide if possible. Potentially the issue could relate to a problem within your OpenDRIVE - but I admit it's a long shot :)
Let's sort out this before we choose whether to pull both or only first commit.

brifsttar:
Hi Emil,

I've sent you my OpenDRIVE. Here are some coordinates to reproduce the bug

```
		track_id_	9
		s_	12.886529064270508
		t_	5.4987192052669789
		lane_id_	4
		offset_	-0.13628079473302090
		h_road_	0.29736141586879061
		h_offset_	0.00000000000000000
		h_relative_	6.2686492420722093
```

From there you can call MoveAlongS(-1), and in my case, I see the method going back and forth between the two connected roads. I now realize that I have modified the enum ```Junction::JunctionStrategyType``` as well as ```MoveToConnectingRoad()``` to give us more control on which road to follow (e.g. right, left), but that shouldn't have any impact here as there's only one connecting lane in the junction.

brifsttar:
Hi Emil,

My colleague just encountered an issue with MoveAlongS on a test drive with a car. I have not been able to check the source of the bug yet, so until I have time to look at it, I would advise against merging the second patch. Sorry for that!

eknabevcc:
Hi Richard. Thanks for info. On the other hand I have, with your ODR, managed to reproduce your issue. And it sure seems to be esmini, and not the ODR. So whatever it will be we need to fix it somehow :) I will investigate further.

eknabevcc:
The plan is to accept the commits. Then I have an additional commit I will try to make on top of this PR. Not being a github pro this may fail, in which case I'll just make the commit afterwards on master.

PR #55, opened 2020-11-27
typo fixes in the xosc examples

johannesWen:


eknabevcc:
Thanks for this cleaning work! 
The elements/attributes are not used in esmini yet, why we've failed to discover the typos ourselves.

PR #62, opened 2020-12-08
Permissive flag and w4 fixes

egunnarsson:
Added /permissive- flag when compiling with Visual Studio and fixed some /W4 warnings.
- Removed unused function copyStateToScenarioGateway.
- Removed unused variables.
- Fixed use of potentially uninitialized variables.
- Fixed bug in OSCAction.hpp.
- Fixed a few "hiding of variable" warnings (OSCPrivateAction.hpp:1155, ScenarioEngine.cpp:472, viewer.cpp:1057)

I would suggest adding the /W4 compile flag as well, there aren't that many warnings overall, so it should be possible to fix all of them. Many of them are C4100 "unreferenced formal parameter" which I find a bit useless, that could be disabled manually with /wd4100, or easily fixed by not naming the unreferenced parameter.

### DIFF
--- a/EnvironmentSimulator/Modules/CommonMini/CommonMini.cpp
+++ b/EnvironmentSimulator/Modules/CommonMini/CommonMini.cpp
@@ -235,8 +235,6 @@ double PointToLineDistance2DSigned(double px, double py, double lx0, double ly0,
 {
 	double l0x = lx1 - lx0;
 	double l0y = ly1 - ly0;
-	double l1x = px - lx0;
-	double l1y = py - ly0;
 	double cp = GetCrossProduct2D(lx1 - lx0, ly1 - ly0, px - lx0, py - ly0);
 	double l0Length = sqrt(l0x * l0x + l0y * l0y);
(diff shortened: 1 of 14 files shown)


EnvironmentSimulator/Modules/CommonMini/CommonMini.cpp
```diff
@@ -235,8 +235,6 @@ double PointToLineDistance2DSigned(double px, double py, double lx0, double ly0,
 {
 	double l0x = lx1 - lx0;
 	double l0y = ly1 - ly0;
-	double l1x = px - lx0;
```
eknabevcc, line 238: I see now that l0x and l0y could have been used in the cross product call. But that's an old mistake, and not a big deal.

eknabevcc (approved):
Excellent cleaning and fixes!

eknabevcc:
I've now added W4 (8295436). As you said, not very much changes needed. But actually at least one potential bad bug found. Thanks a lot for initiating this! 👍

PR #65, opened 2020-12-23
Fixing the broken path caused after files restructuring

ebadi:
I followed the instruction in the [README.md](https://github.com/esmini/esmini) to build RControlStation scenario editor but faced few build issues since the file below is not updated after [this commit](https://github.com/esmini/esmini/commit/bf53f4f29cbbb5da5d4eeaa8772c15b9063303e7). This pull request only fixes the issue in the esmini side.


eknabevcc (approved):
Looks good

PR #71, opened 2021-01-19
Added support for superelevation (OpenDrive)

cfschaller:
Dear @eknabevcc (and colleagues),

first and foremost, thanks for maintaining this great project! We recently decided to have a closer look at it after hearing high praise for esmini and its support of OpenScenario 1.0.
In that context, we realized that esmini lacked support for (lateral) superelevation. As it works mostly similar to the longitudinal elevationProfile, I went ahead and added support for superelevation based on the existing methodology for elevation in this PR.

Looking forward to your comments and all the best,
Christoph

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -1,4 +1,4 @@
-/* 
+﻿/* 
  * esmini - Environment Simulator Minimalistic 
  * https://github.com/esmini/esmini
  *
@@ -1348,6 +1348,10 @@ Road::~Road()
 	{
 		delete(elevation_profile_[i]);
 	}
+	for (size_t i = 0; i < super_elevation_profile_.size(); i++)
+	{
+		delete(super_elevation_profile_[i]);
+	}
 	for (size_t i=0; i<link_.size(); i++)
 	{
 		delete(link_[i]);
@@ -1418,6 +1422,19 @@ void Road::AddElevation(Elevation *elevation)
 	elevation_profile_.push_back((Elevation*)elevation);
 }
 
+void Road::AddSuperElevation(Elevation *super_elevation)
+{
+	// Adjust last super elevation length
+	if (super_elevation_profile_.size() > 0)
+	{
+		Elevation *e_previous = super_elevation_profile_.back();
+		e_previous->SetLength(super_elevation->GetS() - e_previous->GetS());
+	}
+	super_elevation->SetLength(length_ - super_elevation->GetS());
+
+	super_elevation_profile_.push_back((Elevation*)super_elevation);
+}
+
 Elevation* Road::GetElevation(int idx)
 { 
 	if (idx < 0 || idx >= elevation_profile_.size())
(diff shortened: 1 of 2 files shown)


EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
```diff
@@ -1418,6 +1422,19 @@ void Road::AddElevation(Elevation *elevation)
...
 
+void Road::AddSuperElevation(Elevation *super_elevation)
+{
+	// Adjust last super elevation length
+	if (super_elevation_profile_.size() > 0)
+	{
+		Elevation *e_previous = super_elevation_profile_.back();
+		e_previous->SetLength(super_elevation->GetS() - e_previous->GetS());
+	}
+	super_elevation->SetLength(length_ - super_elevation->GetS());
```
eknabevcc, line 1433: Ah, smart, each added superelevation entry gets length = remaining length (s) of the road. Which is then adjusted when/if another is added. Nice.

EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
```diff
@@ -1685,6 +1712,50 @@ bool Road::GetZAndPitchByS(double s, double *z, double *pitch, int *index)
...
+				// Move to next elevation section
+				super_elevation = GetSuperElevation(++ *index);
+			}
+		}
+		else if (super_elevation && s < super_elevation->GetS())
+		{
+			while (s < super_elevation->GetS() && *index > 0)
+			{
+				// Move to previous elevation section
+				super_elevation = GetSuperElevation(-- *index);
```
eknabevcc, line 1743: I see. The superelevation is a polynomial along s, and roll is constant in t dimension. I could expected support for linear change along s, but I checked the standard and the implementation is correct, only poly3 is used. Which makes sense since it can represent line as well. Good.

EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
```diff
@@ -1685,6 +1712,50 @@ bool Road::GetZAndPitchByS(double s, double *z, double *pitch, int *index)
...
+			}
+		}
+
+		if (super_elevation)
+		{
+			double ds = s - super_elevation->GetS();
+			*roll = super_elevation->poly3_.Evaluate(ds);
```
eknabevcc, line 1750: I see. The superelevation is a polynomial along s, and roll is constant in t dimension. I could expected support for linear change along s, but I checked the standard and the implementation is correct, only poly3 is used. Which makes sense since it can represent line as well. Good.

EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
```diff
@@ -1418,6 +1422,19 @@ void Road::AddElevation(Elevation *elevation)
...
+void Road::AddSuperElevation(Elevation *super_elevation)
+{
+	// Adjust last super elevation length
+	if (super_elevation_profile_.size() > 0)
+	{
+		Elevation *e_previous = super_elevation_profile_.back();
+		e_previous->SetLength(super_elevation->GetS() - e_previous->GetS());
+	}
+	super_elevation->SetLength(length_ - super_elevation->GetS());
+
```
eknabevcc, line 1434: I'm thinking what happens if first record is at s > 0, e.g. s = 100. I guess the range 0 < s < 100 will be undefined. A potential fall back could be to add/insert a dummy superelevation record with a=b=c=d=0 before first user added entry.
But on the other hand, maybe it's better that the user is made aware of the "bug" and add an entry from s=0...

eknabevcc (approved):
Excellence! Thank you so much for this contribution. Superelevation support is highly welcome, and the implementation looks very good. 
Some rough testing indicates it works perfect. I will next minutes commit a few minor fixes to road geometry generator (which generates simple road surface in case of missing scenegraph) and it seems to work fine with superelevation.

Issue #72, opened 2021-01-20
Consideration of laneOffset in calculation of t/t_


cfschaller:
Dear Emil,

first off, thanks for the quick review and merge of our recent PR! :)

After some further experimenting with esmini (which is very enjoyable!), we realized that while the laneOffset from an OpenDrive is parsed and also handled within esmini, it is not considered in the calculation of t.

Concretely, the OpenDrive standard states:

![grafik](https://user-images.githubusercontent.com/76468814/105192101-10052180-5b38-11eb-8b0b-59c4561d3999.png)

Thus, the reference for t=0 should be the "road reference line"/"chord line" (i.e. lateral distance from the lane with id=0 plus the laneOffset) and not the "lane reference line" (distance from the lane with id=0). In esmini, t is always calculated with respect to the lane reference line (which works fine, but may be "technically" incorrect). Nonetheless, adjusting this would be a considerable change, so we did not simply want to start implementing, but rather discuss the matter with you. Is this behavior intended? Would you want to change it?

We are aware that the laneOffset is considered as a "correction to t" in esmini where necessary (e.g. in the Method Track2XYZ) as of now. 
If the current treatment of t and laneOffset is supposed to stay, the laneOffset also has to be corrected for in the calculation of superelevation (when implementing it, I had wrongly assumed the laneOffset would be accounted for in t - sorry!). This would only require changing one line, see Commit https://github.com/cfschaller/esmini/commit/389c366547fc0e20dd15b3eaef683246f72539ae (probably I should create a PR for this anyway?).

What do you think?

Thanks again and all the best,
Christoph

eknabevcc:
Hm. This was tricky. I would say that in esmini lateral position = road reference line + lane offset + t.
So if t = 0 then lateral position is just road ref + lane offset. But maybe I'm missing something here. Anyway, I think you're correct in that we don't distinguish between road reference line and lane reference line. E.g. when plotting OpenDRIVE features (toggle key 'o') the center line (thick red) is following any lane offset. Probably it would be more correct to keep it at road center and considering lane reference a different thing, plotted in different color.

Our approach is most often to add and change when needed. So if you (and we) don't have an immediate issue with current implementation I would stick with it for now. But we should keep this in mind for future refactorizations in that code. 

You're solution seems like a very elegant one, aligned with how esmini deal with laneOffset and t, wrong or right :). So yes, please make a PR and I will merge i asap.

cfschaller:
Hey Emil,
[...]

PR #73, opened 2021-01-21
Superelevation support: Fix for non-consideration of laneOffset in t

cfschaller:
Dear Emil,

as discussed in https://github.com/esmini/esmini/issues/72, this is the PR that corrects the superelevation for laneOffset.

All the best,
Christoph


eknabevcc (approved):
Elegant fix!

PR #85, opened 2021-03-09
Bugfix - avoid "faulty" relative hpr

cfschaller:
Hi Emil,

working with the latest 2.4.4 release, I believe to have found the following bug:

1. Using an XOSC with TeleportAction/WorldPosition, SetInertiaPos is called from OSCPositionWorld.
2. SetInertiaPos calls SetHeading/SetPitch/SetRoll
3. In SetHeading/SetPitch/SetRoll not only the values for h_/p_/r_ are calculated but also relative values compared to the road (such as h_relative_)
4. However, h_road_, r_road_, p_road_ are still 0.0 at this point - leading to faulty values being set for the relative h/p/r

Fix: I'd suggest to just set h_, p_ and r_ directly as in the PR.

Update: This PR seems fine now.

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -6257,14 +6257,17 @@ int Position::SetInertiaPos(double x, double y, double z, double h, double p, do
 	x_ = x;
 	y_ = y;
 	z_ = z;
-	h_ = h;
-	p_ = p;
-	r_ = r;
 
 	if (updateTrackPos)
 	{
 		XYZ2Track(false);
 	}
+
+	// Now when road orientation is known, call functions for 
+	// updating angels both absolute and relative the road 
+	SetHeading(h);
+	SetPitch(p);
+	SetRoll(r);
 
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


eknabevcc:
I see now that the tests are all green. So please test this modification and let me know if it still solves your problem. In that case I'll merge it asap.

cfschaller:
Hi Emil,

yes, you are right, the relative angles should still be updated after the road geometry is known  - seems I was a little too fast on the trigger, here.
I tested your modified version and it works fine for me. Looking forward to esmini 2.4.5. :)

EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
```diff
@@ -6257,29 +6257,35 @@ int Position::SetInertiaPos(double x, double y, double z, double h, double p, do
...
-	p_ = p;
-	r_ = r;
 
 	if (updateTrackPos)
 	{
 		XYZ2Track(false);
 	}
 
+	// Now when road orientation is known, call functions for 
+	// updating angels both absolute and relative the road
```
cfschaller, line 6267: Typo "angels" --> angles

EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
```diff
@@ -6257,29 +6257,35 @@ int Position::SetInertiaPos(double x, double y, double z, double h, double p, do
...
 	y_ = y;
-	SetHeading(h);
 
 	if (updateTrackPos)
 	{
 		XYZ2Track(true);
 	}
 
+	// Now when road orientation is known, call functions for 
+	// updating angels both absolute and relative the road
```
cfschaller, line 6286: See above, "angels" --> angles

eknabevcc:
Excellent! Thanks for test and review.

PR #110, opened 2021-05-03
Bugfix - LaneChange over end-to-end/start-to-start road succession

cfschaller:
Hi Emil,

once again - we think to have discovered a bug, offering a "quick fix" with the following PR. :)

Bug description: When performing a lane change while transitioning over a "end-to-end" transition between two roads (e.g. OpenDRIVE "<successor elementType="road" elementId="ID" contactPoint="end" />), the vehicle will jump to the opposite site of the reference line (continuing to drive on the oncoming lane). This is because the target lane (or target_t, respectively) is not corrected for the switch in direction of the t-axis at such transition.

I am not sure whether the offered solution is sufficient - maybe there is a better way to detect the transition then checking whether the t-value changed sign?!

All the best,
Christoph

### DIFF
--- a/EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCPrivateAction.cpp
+++ b/EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCPrivateAction.cpp
@@ -313,7 +313,7 @@ void LatLaneChangeAction::Step(double dt, double)
 	double angle = 0;
 
 	// Detect whether t switched sign due to end-to-end/start-to-start succession of roads
-	if (fabs(t_old + object_->pos_.GetT()) < SMALL_NUMBER && fabs(t_old) > SMALL_NUMBER)
+	if (SIGN(object_->pos_.GetT()) != SIGN(t_old) && fabs(t_old + object_->pos_.GetT()) < SMALL_NUMBER)
 	{
 		start_t_ *= -1;
 		target_t_ *= -1;
--- a/EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCPrivateAction.cpp
+++ b/EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCPrivateAction.cpp
@@ -312,6 +312,14 @@ void LatLaneChangeAction::Step(double dt, double)
 	double factor;
 	double angle = 0;
 
+	// Detect whether t switched sign due to end-to-end/start-to-start succession of roads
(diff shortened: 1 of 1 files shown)


EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCPrivateAction.cpp
```diff
@@ -312,6 +312,14 @@ void LatLaneChangeAction::Step(double dt, double)
 	double factor;
 	double angle = 0;
 
+	// Detect whether t switched sign due to end-to-end/start-to-start succession of roads
+	if (fabs(t_old + object_->pos_.GetT()) < SMALL_NUMBER && fabs(t_old) > SMALL_NUMBER)
```
eknabevcc, line 316: I guess that the purpose of the check `fabs(t_old) > SMALL_NUMBER` is to avoid false cases close to ref line (where the sum may be less than SMALL_NUMBER in spite of same sign). 

This will probably work for 99.999% of all cases. However, if it happens that the road switch happens exactly when t is close to 0, then it might not be detected.

So, to cover all cases, what do you think of making this slight change to the statement:
`if (SIGN(object_->pos_.GetT()) != SIGN(t_old) && fabs(t_old + object_->pos_.GetT()) < SMALL_NUMBER)`
?

eknabevcc (commented):
Excellent finding and a lean solution. I just have one comment and potential minor adjustment, see code below.

EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCPrivateAction.cpp
cfschaller: Yes, that was my intention - but I agree, your suggestion is much better and should capture 100% instead of 99%..

eknabevcc: Great. Would you like to do the change?

cfschaller: Sure, I can do that. :)
Thanks for your feedback!

eknabevcc: Perfect!

eknabevcc:
Thanks for this contribution, really an improvement. 👍

cfschaller:
Thanks for the collaboration, as always - it was a pleasure! 👍

PR #141, opened 2021-08-07
Added some const correctness to RoadManager

brifsttar:
Hi Emil,

In my own modules, I tend to be somewhat strict about const-correctness. Since I use RoadManager a lot, over time I had to propagate some of my const-correctness to it. It's obviously not complete, but it's a starting point.

I'm not sure where you stand on const-correctness, since it has its pros and cons. But I thought I'd make a PR, that way you can decide and I'll know :-)

Thanks again for all your work.

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -6914,7 +6914,7 @@ double Position::GetCurvature()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


eknabevcc:
Excellent. I've been lazy. I'm not that aware, but to me it really makes sense to indicate and establish whether a method is expected or allowed to make changes to the "this" object. Hopefully no one else will object, we'll see... 
Thanks for contribution!

PR #151, opened 2021-09-02
Removing ghost vehicle info from OSI groundtruth dynamic

ZijianHan-VolvoCars:
Hello, it seems the checks are passed.


ZijianHan-VolvoCars:
Please dont merge it now because we find out there are some test using ghost vehicle data in osi

PR #162, opened 2021-09-29
reference driver model from ECE ALKS regulation

tmIdiada:
Please merge the new controller. Already tested, if it's working like expected.

Whenever you want, you can delete the LOG messages, but for me they would be important to understand the driver model within the corresponding scenario.

I've sent this pill request into alks instead of master, that you can better understand the changes, but the master would be even better.

Many thanks for your support, especially for the new methods like GetVelT.

If you want I can create another pull request directly into master.

### DIFF
--- a/EnvironmentSimulator/Modules/Controllers/ControllerECE_ALKS_DRIVER.cpp
+++ b/EnvironmentSimulator/Modules/Controllers/ControllerECE_ALKS_DRIVER.cpp
@@ -11,7 +11,10 @@
  */
 
 /*
- * This controller simulates the driver from ECE ALKS regulation.
+ * Written by: Thaddaeus Menzel, IDIADA Fahrzeugtechnik GmbH
+ * This controller simulates the reference driver model from ECE ALKS regulation.
+ * For more information look at:
+ * https://unece.org/sites/default/files/2021-03/R157e.pdf, pages 43-45
  */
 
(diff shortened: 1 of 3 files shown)


eknabevcc (approved):
Very good! Let's start merging into the alks branch. 
I can see two improvements before merging into master:
1. Default no logging. But add a controller property so user can activate logging, e.g. "logging" true/false. 
2. Replace static variables with class member variables. 
Here I need to apologize. It was my initial suggestion to use static variables for keeping track of elapsed time. But I realize that this controller might be added to multiple vehicles, in which case they would share static variables - no good. 
Your first solution to have them as class members was better. But for some reason you changed into static ones, maybe according to my bad proposal.
I simply suggest to move them back as class members.

Anyway, let's merge now. And we can fix the above two items as next step.

eknabevcc:
I now pushed commit 6d6f648edf11854f30a502177160e0229b2347dc which solves:
1. Make logging optional (off by default). Add this line to scenario file:
                    `<Property name="logging" value="true" />`
  to activate logging. For example:
```
           <ObjectController>
            <Controller name="ECE_ALKS_DRIVER_Controller">
                <Properties>
                    <Property name="timeGap" value="1.0" />
                    <Property name="logging" value="true" />
                </Properties>
            </Controller>            
         </ObjectController>  
```
2. Moved the static variables into the class as normal member variables. This way we can assign ALKS controller to many vehicles in parallel 👍

PR #167, opened 2021-10-04
Fix - no more alternating offsets for AcquirePosition

cfschaller:
Dear Emil,

when using the "new" AcquirePositionAction, we observed that for some cases (depending on the route direction) the lane offset of the underlying vehicle would be alternated (in terms of sign) for each simulation step.

This PR contains a potential fix.

Looking forward to your feedback and once again thanks for your efforts,
Christoph


eknabevcc (commented):
Excellent finding, we were not aware of this issue.
Your solution is a good step. However I suspect it will not respect if offset is changed after initialization. Please see and test suggested/committed changes. Does it still work with your use case?

PR #177, opened 2021-10-15
Alks

tmIdiada:
First final version (could be merged into master, will be no more changed so soon). All scenarios of the ECE ALKS regulation have been simulated and the 'same' plots have been generated.

There are for sure some differences to the regulation, but these changes are mainly related to the differences in objects' width

cut-out with dv=10kph, vy=0.1m/s and ds=15m (free space) causes no leading vehicles collision:
vy=0.1m/s is the maximal lateral velocity, but assume it is constant:
TTC = ds / dv = 5.4s
dy = TTC * vy = 0.54m -> cut-out object and object in front of it must have a width smaller than 0.54m and I've taken here already ds=15m, if one looks at the plot, then already a ds=13m causes no leading vehicles collision (width < 0.468m). And if one also takes vy=0.1m/s as maximal lateral velocity, not constant one, then the objects must be even smaller (unrealistic, since in EUNCAP AEB test protocol a bicyclist has already a width of 0.5m)

### DIFF
--- a/resources/xosc/alks-test.xosc
+++ b/resources/xosc/alks-test.xosc
@@ -39,11 +39,7 @@
       <ScenarioObject name="Ego">
         <CatalogReference catalogName="VehicleCatalog" entryName="$HostVehicle"/>
          <ObjectController>
-            <Controller name="ECE_ALKS_DRIVER_Controller">
-                <Properties>
-                    <Property name="timeGap" value="1.0" />
-                </Properties>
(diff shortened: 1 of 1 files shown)


eknabevcc:
Thanks. Looks good. Let's merge this on the alks branch. Next steps is to rebase and merge on master. Will try to do that coming week.
I'm thinking that we could call this controller ALKSRefController, what do you think?

PR #194, opened 2021-12-07
adding some variables to global scope

ebadi:
This pull request makes it possible to use esmini's cmake variables by changing their scope (making it visible to the top level parent cmake scripts). The goal  is to be able to build esmini as a sub module using `add_subdirectory(${esmini_SOURCE_DIR})`.

Reference:
https://cmake.org/pipermail/cmake/2010-July/037943.html


eknabevcc:
Nice contribution, thanks!

PR #217, opened 2022-01-24
Some changes in ECE ALKS reference driver

tmIdiada:
Like promised better later/late than never, here the newest version of the reference driver, which was the baseline for the system under test. All simulation have been executed from the regulation with this version of the reference driver. European funded HEADSTART project finished end of last year, so no further changes from our side expected for longer time.

Changes due to corner cases, e.g. in case of narrow lanes, high lateral velocities or large timesteps.


eknabevcc:
Thanks for the contribution!
It's now merged on master and will be included in next release.

PR #230, opened 2022-02-08
Fix loadopendrivethatdonothaveallroadsandjunctioninonefile

eriadam36:
simple solution to difficult segmentation fault error, when loading a part of az opendrivemap...


eknabevcc:
Sorry, I must have missed this PR. 
It looks as good stuff. What's the reason for deleting the branch, is it not relevant anymore?
Or could it be that other commits (we did some clean up of nullptr references in OSI/road manager recently) solved same issue?

PR #244, opened 2022-02-24
some corrections and added script to install all for esmini needed libs at once

tmIdiada:


eknabevcc:
This is great improvements to the scripts! And nice with the combo script that builds all libs at once. Good work. 👍
I did minor tweak to support starting script from any folder. 
A second commit was to cover a special case in MSYS where the script is called by absolute path, e.g: `/c/tmp/esmini/scripts/generate_osi_sumo_osg_libs.sh`
I hope you don't mind the commits on your branch. In that case we can move to a separate branch.
Let me know if you see any issues. Otherwise I suggest we go ahead and merge.

eknabevcc:
This is great improvements to the scripts! And nice with the combo script that builds all libs at once. Good work. 👍
I did minor tweak to support starting script from any folder. 
A second commit was to cover a special case in MSYS where the script is called by absolute path, e.g: `/c/tmp/esmini/scripts/generate_osi_sumo_osg_libs.sh`
I hope you don't mind the commits on your branch. In that case we can move to a separate branch.
Let me know if you see any issues. Otherwise I suggest we go ahead and merge.

tmIdiada:
All mods ok to me, good job from your side. We can go ahead

scripts/generate_osi_sumo_osg_libs.sh
```diff
@@ -0,0 +1,105 @@
...
+    fi
+    if [[ ! -d "$SUMO_BUILD_DIR/$BUILD_SUB_DIR" ]]; then
+    	mkdir "$SUMO_BUILD_DIR/$BUILD_SUB_DIR"
+    fi
+    cd "$SUMO_BUILD_DIR/$BUILD_SUB_DIR"
+    $SUMO_BUILD_SCRIPT
+    cd ..
+    "$z_exe" x "$BUILD_SUB_DIR/sumo_$target_dir.7z"
+    if [ "$OSTYPE" == "msys" ]; then
+        rmdir "$BUILD_SUB_DIR" /s /q
```
eknabevcc, line 100: This fails in git bash on Win10. What do you think of changing into
`rm -rf "$BUILD_SUB_DIR"`
?
Actually, then we can skip this if-statement altogether. What do you think?

tmIdiada: This is the difference between Win cmd snd powershell? Thought rm -rf does not work for win cmd, are you using power shell? But for sure if rm -rf works for you, you can delete the if condition, have tested only on linux, rest searched via google

eknabevcc: I see. I was lazy enough to limit the Windows case to only support (Git) bash. I think msys/msys2 bash would work equally good (guess it's the same). 

I agree that it would be complex to support all possible command interpreters (PowerShell, cmd, bash...). So I think it's perfectly OK to stick with bash support for now. At least we should postpone hard work until it's needed :)

I found and fixed a few flaws in the OSG script on Windows that was introduced by myself in last update. 

Now everything seems to compile perfectly on Windows as well - in one single command! Really nice 👍

tmIdiada: ok, then please remove the if conditions if not already done

eknabevcc: I will, and then I think we're ready for merge

eknabevcc:
I had to resort to command line to clean up (squash my multiple commits) and then rebased and merged on master. 
I thought it would be indicated here as well but it didn't. 
Anyway, your commit is in and visible (32f8f1721238be46bccf2d4e0dcc50231d402f39)

Thanks a lot for contribution!

I close this now manually.

PR #247, opened 2022-02-28
Fix - AcquirePositionAction

cfschaller:
Hi Emil,
we were observing some problems using the AcquirePositionAction lately, in particular, if the target position was on the same road(ID) as the initial position. In that case, the vehicle was either freezing at the end of the road or teleporting back to its start.
Also, for some TargetPositions the routing never stopped (even though the vehicle passed the target position), because the sign of the route direction was not determined correctly. Those cases are easily reconstructed using simple AcquirePositionActions on the aforementioned Town01.xodr. If necessary, I can also upload example XOSCs.

This commit should fix the aforementioned issues (related to determining route direction), via
- Usage of another waypoint from all_waypoints_ when there is only one minimal waypoint
- Comparison of PI/2 < heading < 3*PI/2 (instead of 3*PI/4 on the right hand side) to determine the (heading-based) direction
- Returning the heading-based direction as the default (which is necessary to properly keep track of how far on the route a vehicle has already moved)
- Warning if route direction and heading direction are not aligned

Please have a look whether I missed anything or broke some other functionality by accident.

Thanks in advance,
Christoph

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -10222,7 +10222,7 @@ Position::ErrorCode Route::MovePathDS(double ds)
 	}
 
 	// Consider route direction
-	ds *= GetDirectionRelativeRoad();
+	ds *= GetWayPointDirection(waypoint_idx_);
 
 	return SetPathS(GetPathS() + ds);
 }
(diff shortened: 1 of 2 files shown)


eknabevcc:
Excellent findings and thanks for providing a solution to the problem as well. Appreciated!
It's now merged on master.

PR #248, opened 2022-03-02
Fix - linear LaneOffset Action should properly use the given rate

cfschaller:
Hi Emil,

yet another pull request - hoping that our change finds your approval and is in line with the OpenSCENARIO standard.

Using a "linear" LaneOffset action we were wondering that esmini calculated other durations for the respective maneuvers than we would have expected. For example, for a vehicle starting in the lane center and targeting an absolute lane offset of 1m, we would have expected the maneuver to take 2 seconds when giving it a rate of 0.5 (=what we understand as linear behavior).

Looking at the current esmini-code, we revisited the standard but still understand that the given parameter for the LaneOffsetAction should be used as the rate of the linear transition, which is achieved by the suggested change.

### DIFF
--- a/EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCPrivateAction.cpp
+++ b/EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCPrivateAction.cpp
@@ -117,9 +117,8 @@ double OSCPrivateAction::TransitionDynamics::GetTargetParamValByPrimPrimPeak(dou
 	}
 	else if (shape_ == DynamicsShape::LINEAR)
 	{
-		// Acceleration is infinite at start and end, anything else should result in flat line
-		// Just to have something reasonable, re-use acc equation from CUBIC case
-		return sqrt(6 * abs(GetTargetVal() - GetStartVal()) / prim_prim_peak);
+		// For the linear case, the duration should be distance / rate.
+		return abs(GetTargetVal() - GetStartVal()) / prim_prim_peak;
 	}
 	else if (shape_ == DynamicsShape::SINUSOIDAL)
 	{


eknabevcc:
To my understanding (unfortunately) the LaneOffset action is not using rate in the same way LaneChange do. 

Rate (change rate of lateral movement) is an attribute of the LaneChange [dynamicsDimension](https://www.asam.net/static_downloads/ASAM_OpenSCENARIO_V1.1.1_Model_Documentation/modelDocumentation/content/DynamicsDimension.html). 
While in LaneOffset the user specifies maximum lateral acceleration (maximum change rate of lateral speed) as part of [LaneOffsetActionDynamics](https://www.asam.net/static_downloads/ASAM_OpenSCENARIO_V1.1.1_Model_Documentation/modelDocumentation/content/LaneOffsetActionDynamics.html).

Maximum lateral acceleration is defined for (C1) continuous shapes like sinusoidal and cubic. However, for the linear shape the lateral acceleration is infinite at start and endpoints (where the motions changes from and to moving straight along the lane). So, to have any reasonable mapping of the maxLateralAccleration attribute we simply picked same calculation as for the cubic case.

This difference between the nearly related lane actions in OpenSCENARIO has been and still is under discussion. One reason for keeping the differences, instead of aligning, is that most often they can solve the same use case while providing different interface in effect providing richer use case coverage.

So, in your case, if you need/want to specify lateral rate, would it be possible to define the action in terms of LaneChange action instead of LaneOffset, e.g. keep lane but change offset? If not, please explain the limitation, it can be useful input.

cfschaller:
Hi Emil,

first off, sorry for the delayed reply!
I agree that for the LaneOffsetAction the maximum lateral acceleration (and not the change rate as for the LaneChange) is to be provided. However, as you stated, that does not make sense for the linear shape where there is infinite accelerations at the start/end and zero otherwise.

Therefore, I would still see the following argumentation for using the suggested PR instead of the formula for the cubic case:
- the calculation for the cubic case does not have anything to do with the linear shape of the action - it is not clear for the user what kind of linear movement will result from providing a certain rate (or is there really a use case where the user would want to provide the cubic rate for the linear shape?)
- in contrast, providing the linear rate is intuitive (as it is the one value describing the linear shape)

What do you think?

Finally: Yes, a LaneChangeAction with the same lane and the target offset can be used as a replacement for our use case.

cfschaller:
Hi Emil,

I see your point. We will use the LaneChangeAction for now and I will get back to you in case we come accross a limitation.

Thanks again for all the details - I will close this PR for now.

PR #249, opened 2022-03-02
Suggestion: new esminiLib-Method to register a callback for triggered conditions

cfschaller:
Hi Emil,

and another one - but this time it is not a fix, but a suggestion for a new feature.

UseCase: We would like to know from a simulation run in esmini when a certain condition was triggered. Of course, it would be possible to parse the log/console output, but a much cleaner solution seems to be registering a callback that is called every time a condition is triggered.

What do you think?

Feel free to ask further questions, request changes or make suggestions - we want to make this an improvement of esmini and not just a function we "need".

### DIFF
--- a/EnvironmentSimulator/Libraries/esminiLib/esminiLib.cpp
+++ b/EnvironmentSimulator/Libraries/esminiLib/esminiLib.cpp
@@ -22,6 +22,7 @@
 #endif
 #include "vehicle.hpp"
 #include "pugixml.hpp"
+#include "OSCCondition.hpp"
 
 using namespace scenarioengine;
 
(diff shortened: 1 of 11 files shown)


cfschaller:
Hi Emil,

and a little update on this one. We added a second callback that can be registered: this time not for conditions, but for events (where there is event start and event end, which is signaled by a bool in the callback).

Looking forward to your feedback!

eknabevcc:
Great contribution! I like these callback features. I think the can be really useful.

I did a manual rebase, solved one minor conflict, and finally merged on master.
Thanks! Closing the PR.

PR #250, opened 2022-03-03
libjpeg upgrade to newest release 9e, some fixes

tmIdiada:
- newest release libjpeg9e supported, tested under linux x86_64 architecture with ubuntu18.04 in graphical mode, additionally build tested on aarch64 linux architecture (visual result could not be tested due to some graphical issues, nvidia tegra graphics chip), aarch64 linux architecture supported since version libjpeg-9, but until actual version 9e even more architectures are supported and no issues could be identified using the newest version here
- some fixes have been done with variables/paths
- fbx per default activated now on linux and mac, but I had problems here on linux, not sure if this installation really works, I had this variable always per default as false

### DIFF
--- a/scripts/generate_osg_libs.sh
+++ b/scripts/generate_osg_libs.sh
@@ -28,6 +28,8 @@
 # -----------------------------------------------------------------------------------
 # Review and update settings in this section according to your system and preferences
 
+fbx_support=false  # users are encouraged to convert fbx to osgb format whenever possible
+
 if [ "$OSTYPE" == "msys" ]; then
 	# Visual Studio 2019 - toolkit from Visual Studio 2017
 	GENERATOR=("Visual Studio 16 2019")
@@ -53,8 +55,6 @@ if [ "$OSTYPE" == "msys" ]; then
     target_dir="v10"
     zfilename="osg_v10.7z"
     z_exe="$PROGRAMFILES/7-Zip/7z"
-    fbx_support=false
-
 
 elif [[ "$OSTYPE" == "darwin"* ]] || [[ "$OSTYPE" == "linux-gnu"* ]]; then
 	# Unix Makefiles (for Ubuntu and other Linux systems)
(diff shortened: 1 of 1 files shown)


eknabevcc:
I will look at the FBX issue(s). 
Actually, one reason for bringing in FBX was to support animations in future.
However, I recently tested to convert some 3D FBX files with embedded animations to OSGB. And to my positive surprise the animations followed and worked just as fine in the .osgb file. So I'm now leaning towards skipping plans for FBX support by default in esmini, and instead promoting conversion of such files to osgb. Main reason to avoid esmini binaries grow. FBX is a huge library, especially the debug version. 

So my current thinking is to remove FBX support from the osg libs built for esmini. 
Instead encourage and show how to convert into osgb. 
What do you think?

tmIdiada:
> I will look at the FBX issue(s). Actually, one reason for bringing in FBX was to support animations in future. However, I recently tested to convert some 3D FBX files with embedded animations to OSGB. And to my positive surprise the animations followed and worked just as fine in the .osgb file. So I'm now leaning towards skipping plans for FBX support by default in esmini, and instead promoting conversion of such files to osgb. Main reason to avoid esmini binaries grow. FBX is a huge library, especially the debug version.
> 
> So my current thinking is to remove FBX support from the osg libs built for esmini. Instead encourage and show how to convert into osgb. What do you think?

Hi, yes I like the idea to keep esmini as slim as possible, especially regarding external libs, like sumo, osi, osg, fbx, libjpeg, ... the less the better it is and easier for updates/upgrades. Then you should set the variable to false value also for linux and mac build per default.

eknabevcc:
I tried to remember whether the choice of jpeg 8d was on some purpose, instead of going for the latest. Anyway, I can't remember it now, and it seems to build just fine on Kubuntu and Ubuntu. So let's go for it!
Thanks again for your efforts. Great that esmini portability is expanding 👍

PR #265, opened 2022-03-30
Fixed issue in RoadPath::Calculate

brifsttar:
Hi Emil, it's me again

A colleague and I tracked down an issue, still related to `Position::Delta`. This time, we sometimes had a `ds` with the wrong sign. Our investigation lead us there:

https://github.com/esmini/esmini/blob/45ef22ed04867f6e6630bb73461e5d4a3fcd7d93/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp#L4510-L4520

We were in the first cast of the `OR` (`node->link == startRoad->GetLink(LinkType::PREDECESSOR)`), and our `HRelative` was like 2*PI, so we fully expected the condition to be false, to have `direction_` set to `-1` and then to have proper signing on our `ds`.

However, the second part of the `OR` seems to be missing a set of parenthesis, which meant the condition was evaluated as `A AND B OR C` instead of what we're assuming should be `A AND (B OR C)`.

In our case, A was obviously false, but C was true. And apparently, in C++, `false && false || true` is `true`, so we got in the condition, `direction_` was set to `1`, which lead to our problem.

We fixed that pretty hastily by just adding parenthesis, but I think we should have taken this opportunity to rewrite this condition more clearly, which makes it both more readable and help reduce the risk of mistakes. Something like the following

```cpp
bool isPred = node->link == startRoad->GetLink(LinkType::PREDECESSOR);
bool isGTPi2 = abs(startPos_->GetHRelative()) > M_PI_2;
bool isLT3Pi2 = abs(startPos_->GetHRelative()) < 3 * M_PI / 2);
bool isSucc = node->link == startRoad->GetLink(LinkType::SUCCESSOR);
bool isLTPi2 = abs(startPos_->GetHRelative()) < M_PI_2;
bool isGT3Pi2 = abs(startPos_->GetHRelative()) > 3 * M_PI / 2;
if ((isPred && isGTPi2  && isLT3Pi2) || (isSucc && (isLTPi2  || isGT3Pi2 ))) {
    // ...
}

// or

bool isPredAndBack = isPred && isGTPi2  && isLT3Pi2;
bool isSuccAndFront = isSucc && (isLTPi2  || isGT3Pi2 );

if (isPredAndBack || isSuccAndFront) {
    // ...
}

```

The compiler will optimize all that anyway, so it should have no performance impact.

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -4510,7 +4510,7 @@ int RoadPath::Calculate(double &dist, bool bothDirections, double maxDist)
 					if ((node->link == startRoad->GetLink(LinkType::PREDECESSOR) &&
 						abs(startPos_->GetHRelative()) > M_PI_2 && abs(startPos_->GetHRelative()) < 3 * M_PI / 2) ||
 						((node->link == startRoad->GetLink(LinkType::SUCCESSOR) &&
-						abs(startPos_->GetHRelative()) < M_PI_2 || abs(startPos_->GetHRelative()) > 3 * M_PI / 2)))
+						(abs(startPos_->GetHRelative()) < M_PI_2 || abs(startPos_->GetHRelative()) > 3 * M_PI / 2))))
 					{
 						direction_ = 1;
 					}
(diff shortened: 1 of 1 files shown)


esmini:
Excellent finding! Yes I now also see the error, by pure visual inspection of the if-case. Sometimes I add obsolete parenthesis around AND components even though they should have precedence over OR. But sometimes I skip and rely on the rules. In this case I seem to mixed the two strategies 🙃

I was about to accept the PR immediately. But I find your further proposed code clarification beautiful. Can I ask you to commit that on this PR? In that case I keep it open until then.

eknabevcc:
Very good. Let's start merging this. 👍
I'll look into the other issue when I can find time.

eknabevcc:
The initial issue is fixed and included in release [2.22.0](https://github.com/esmini/esmini/releases/tag/v2.22.0).

PR #266, opened 2022-04-01
[roadmngr] Fixed Position::Delta sometimes flipping dLandId

### LINKED ISSUE #261: Position::Delta sometimes return wrong dLaneId
Hi Emil,

I hope you're doing well. It's been a while since I found a bug in esmini, but it's not for lack of trying! We've been using RoadManager more and more recently, and I can't thank you enough for your work on this project.

Anyway, we recently encountered an issue with `Position::Delta`, which returned an incorrect `dLaneId`. As usual, I brought my A-game MSPaint skills to help illustrate the issue.

image

So the issue occurs when requesting the delta between points `A` and `B`. They are indeed connected via the junction `j1`, and their `dLanedId` should be either 2 or -2 (I can never remember which side is which). But `Position::Delta` returns a `dLanedId` of 0, which isn't what I expected.

I've tracked down the issue to the following lines, and I'll explain more about it below.

https://github.com/esmini/esmini/blob/45ef22ed04867f6e6630bb73461e5d4a3fcd7d93/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp#L8272-L8277

Basically, Dijkstra's algorithm works as expected, and correctly finds the connection between `r1` and `r2` via `j1`. However, when navigating the graph of connected roads from `r1`, it starts first by checking `j2`. Which is not an issue, as the algorithm properly backtracks to `j1` and finds the connection.

However... `j2` gets added to the `visited_` vector, which is where trouble starts. That, in itself, obviously isn't an issue, but it leads to one down the line.

https://github.com/esmini/esmini/blob/45ef22ed04867f6e6630bb73461e5d4a3fcd7d93/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp#L4494

Because if we go back to culprit lines above, we see that you check if start and end road are inverted to (rightfully) flip sides. However, the code assumes `visited_[0]` as the first `RoadLink` between the two points, but that's not always the case! And in my example, it's definitely not the case, as the first checked `RoadLink` happened to be the `j2` connection, which wasn't the correct one. The consequence is that the code thinks roads are in opposite directions, so it flips sides, which leads to a null `dLanedId` when I expected 2 (or -2, I still don't know :P ). If point `B` was on the other lane of `r2`, `dLanedId` would then be 2 instead of 0.

If my analysis is correct (and if I managed to make it understandable), one solution could be to add another attribute to `RoadPath` of type `std::vector<PathNode*>` which would store the actual solved path between the two points. You could then use that variable instead of `visited_` in the code above, and that would work. If you want, I could probably try to implement that and make a PR, but I still wanted to raise the issue to you first.

If you want, I can also send you the `.xodr` (it's a new one!) along with the actual coordinates leading to the erroneous `Position::Delta` call.

brifsttar:
Hi Emil,

This should fix #261. I've taken this opportunity to also make the `if` condition more readable.

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -8284,11 +8284,11 @@ bool Position::Delta(Position* pos_b, PositionDiff &diff, bool bothDirections, d
 			bool isPathBackward = firstNode->link->GetType() == LinkType::PREDECESSOR;
 			bool isConnectedToEnd = lastNode->link->GetContactPointType() == ContactPointType::CONTACT_POINT_END;
 			bool isConnectedToStart = lastNode->link->GetContactPointType() == ContactPointType::CONTACT_POINT_START;
-			bool isHeadToToe = isPathForward && isConnectedToEnd;
-			bool isToeToHead = isPathBackward && isConnectedToStart;
+			bool isHeadToHead = isPathForward && isConnectedToEnd;
+			bool isToeToToe = isPathBackward && isConnectedToStart;
 
 			// If start and end roads are oppotite directed, inverse one side for delta calculations
-			if (isHeadToToe || isToeToHead) {
+			if (isHeadToHead || isToeToToe) {
 				laneIdB = -laneIdB;
 				tB = -tB;
 			}
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -4529,13 +4529,14 @@ int RoadPath::Calculate(double &dist, bool bothDirections, double maxDist)
 		}
 	}
 
-	// Compensate for heading of the start position
-	if (startPos_->GetHRelativeDrivingDirection() > M_PI_2 && startPos_->GetHRelativeDrivingDirection() < 3 * M_PI_2)
-	{
-		direction_ *= -1;
-	}
 	dist = direction_ * tmpDist;
 
+	// Also take intial heading of the start position into consideration for the sign of the distance
+	if (startPos_->GetHRelativeDrivingDirection() > M_PI_2 && startPos_->GetHRelativeDrivingDirection() < 3 * M_PI_2)
+	{
+		dist *= -1;
+	}
+	
 	return found ? 0 : -1;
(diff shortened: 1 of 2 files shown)


EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
```diff
@@ -8274,15 +8274,24 @@ bool Position::Delta(Position* pos_b, PositionDiff &diff, bool bothDirections, d
...
-				path->visited_.back()->link->GetContactPointType() == ContactPointType::CONTACT_POINT_END) ||
-			 (path->visited_[0]->link->GetType() == LinkType::PREDECESSOR &&
-				path->visited_.back()->link->GetContactPointType() == ContactPointType::CONTACT_POINT_START)))
-		{
-			laneIdB = -laneIdB;
-			tB = -tB;
+		if (path->visited_.size() > 0) {
+			RoadPath::PathNode* lastNode = path->visited_.back();
+			RoadPath::PathNode* firstNode = lastNode;
+			while (firstNode->previous) {
```
eknabevcc, line 8280: I understand this loop. It seems like a flaw that visited[0] is not the actual first road in the "path".
Still it would be nice to skip the loop. I'll check later if there is another solution (to store FIRST road).

eknabevcc: Yes, I like the idea to store the already calculated start direction. And then re-use it the Delta method which could be used frequently, hence we want to minimize unnecessary work.

I took the liberty of committing an attempt to achieve this on your branch. If it builds, please test and report how it works.

eknabevcc:
Good that a test case caught a bad attempt from my side.

I made another commit that seems to work better. But since it included also changes in RoadManager.hpp (which is not changed in your commit) I can't edit it in this PR. I tried to figure out a way to commit to the PR branch, but I had to give up.

Instead I created a branch [fix_flip](https://github.com/esmini/esmini/tree/fix_flip). Could you perhaps merge that branch on your (PR branch)? 

As a last resort we could close this PR and I could merge fix_flip on master instead.

brifsttar:
I merged your branch on this PR, and I ran the test that initially failed (see #261), and it worked as expected. So all is ok on my end.

eknabevcc:
Excellent, thanks for the fix!

PR #268, opened 2022-04-06
Fix - reset event callback in resetScenario

cfschaller:
Hi Emil,

unfortunately our initial contribution was missing this line to reset the event callback.
If nothing is missing, please merge this.

Thanks!


eknabevcc:
Excellent! Thanks for the fix.

PR #283, opened 2022-05-13
[roadmngr] Fixed Delta sometimes returning wrong dt/dlane in junction

brifsttar:
Hi Emil,

This PR is somewhat related to #261 and #266, as it fixes some errors in `Position::Delta`. As usual, I brought my MSPaint skills to help illustrate the issue.

image

I've also attached the `xodr` file [here](https://github.com/esmini/esmini/files/8685974/delta_junction.txt) (as `txt` because of GitHub restrictions), and the coordinates are as follow

```
A
  X=782.53093750
  Y=1645.91328125
  Road=677
  Lane=1
  S=8.09
  T=-0.517
B
  X=782.96906250
  Y=1656.29312500
  Road=18
  Lane=1
  S=2.141
  T=0
```

So if we call B.Detla(A), we have a `dLaneId` of 0, where we expected 2. The issue lies here:

https://github.com/esmini/esmini/blob/ab19cd2a7e97fc2157d21345d092b3dc078fe73d/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp#L8311-L8312

The roads (18 and 677) are directly connected through the junction (665), so resolving the path is quite straightforward. However, the `lastNode`'s link is the road to junction link, meaning its `ContactPointType` is `CONTACT_POINT_JUNCTION`. So both bools are false, and HeadToHead/ToeToToe checking can't proceed to flip the LaneID and T.

This PR (hopefully!) solves this issue by checking if the last contact point is a junction, and if so, checks the connection by looking at the target road's predecessor and successor. I've tested it on the failing test mentioned above, plus some other random locations on the network, and I encountered no issue. But I didn't do any thorough testing.

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -8310,6 +8310,16 @@ bool Position::Delta(Position* pos_b, PositionDiff &diff, bool bothDirections, d
 			bool isPathBackward = firstNode->link->GetType() == LinkType::PREDECESSOR;
 			bool isConnectedToEnd = lastNode->link->GetContactPointType() == ContactPointType::CONTACT_POINT_END;
 			bool isConnectedToStart = lastNode->link->GetContactPointType() == ContactPointType::CONTACT_POINT_START;
+			bool isConnectedToJunction = lastNode->link->GetContactPointType() == ContactPointType::CONTACT_POINT_JUNCTION;
+			if (isConnectedToJunction) {
+				Road* lastNodeRoad = lastNode->fromRoad;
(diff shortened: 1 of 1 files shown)


eknabevcc:
Please refresh my memory, is this PR still valid? 
Or was it fixed by issues #261 and/or #266?

brifsttar:
Hi Emil,

This PR should still be valid, its relation to #261 and #266 is that it involves issues in `Position::Delta()` that are in the same code section, but those are distinct issues.

The previous PR was to fix an issue where the first node of the calculated road path was wrongly assumed to be the first visited road. This new PR fixes the unhandled case of the calculated road path's last node being connected via a junction contact point, instead of the usual start/end.

I hope it's clearer now.

eknabevcc:
Good. Thanks for the fix!

Issue #289, opened 2022-06-12
Issue of Multi-Actors Executing


gaogaihetao:
**description**
When the actor list in ManeuverGroup has more than 1 Entities, only the last one can execute the inside Action successfully. E.g. when car1, car2, car3 are packed in a ManeuverGroup to execute a SpeedAction, only car3 moves but car1 and car2 keep standstill. Also, this issue won't occur if StartTrigger is omitted.

I found the related [code](https://github.com/esmini/esmini/blob/master/EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCManeuver.cpp#L63): A PrivateAction will be deducted to be conflicting with a running one if they occupy the same domain. Actually, I think it's not so appropriate because it doesn't consider that a PrivateAction might be executed by different actors. I did a test to add a judgement of Actor name in codes, and all Actions worked. Please correct me if I am wrong.

eknabevcc:
I think you're totally right! Nice finding.
I think your approach for solution makes sense. Would you like to do a pull request? 
It would save time for me, and you would get the creds correctly :)

gaogaihetao:
Thanks for your permission! I have started a PR and please take a look.

---issue closed--

PR #290, opened 2022-06-15
Fix multi-actors execution issue #289

gaogaihetao:
- Consider actor's name when deducting conflicting actions


eknabevcc:
Excellent! Thanks for finding this flaw and fixing it. 
I made a small modification, looking at object id instead of name. Do you see any drawback with that change?

gaogaihetao:
No. Thanks for helping fix name. It looks good to me.

PR #292, opened 2022-06-17
Fix Wrong successor selecting at junction

gaogaihetao:
This is a bugfix handling of a route tracking issue.

**Issue**
I drew a picture to illustrate an issue I met:
junction
On the assigned route, if the incoming road has more than 1 link at the junction that links to different lanes of the same outgoing road, it might select a wrong connecting road and finally go to the wrong target lane of the outgoing road.

**Example**
Here is an example of this issue:
j
issue
The car was supposed to enter the rightmost lane as the routing waypoint specified, but it went to the wrong connecting road at the junction.

**Solution**
I think this road type commonly occurs in daily use, and it's not a map making issue. So I tried to find out the reason in esmini's code: When searching a connecting road at a junction, only outgoing road id is checked whether to be accordant with waypoint's road id. It doesn't consider that the outgoing road might have more than 1 accessible lane.

To solve that, it needs to check the lane id of the outgoing road with the lane id of the next waypoint, or check the connecting road id with the tracked road id. I chose the latter method because code modification is very simple.

**Test**
I tested the fixed code on the example: The car can go to the correct road now.
fixed

Would you please take a look?

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -7334,7 +7334,7 @@ Position::ReturnCode Position::MoveToConnectingRoad(RoadLink *road_link, Contact
 					if (connecting_road)
 					{
 						Road* outgoing_road = junction->GetRoadAtOtherEndOfConnectingRoad(connecting_road, road);
-						if (outgoing_road == outgoing_road_target)
+						if (outgoing_road == outgoing_road_target && connecting_road->GetId() == r->GetTrackId())
 						{
 							connection_idx = i;
 						}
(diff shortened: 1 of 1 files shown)


eknabevcc:
Very nice! Good finding and smart solution. 
I see that your second commit relax the routing, allowing way-points that do not connect lane to lane, which is good to continue supporting.

gaogaihetao:
A smoke test of route_lane_change failed when I submitted the 1st commit. (Sorry I forgot to do automated testing locally.) I reviewed the case and found my commit actually changed the original behavior. In that case, LaneChangeAction changes the connection road that will be checked but doesn't change the tracked road id, in which situation the tracked road and the connecting road are not the same. In my first commit, no connection index (default to be 0) will be updated.
So I submit the second commit:
- If there is a connecting road linking to target lane of road, select it; (be referred as the optimal connection)
- If there is no connecting road linking to the target lane, select a connection linking to the target road.
Would you mind telling me if there's a better way?

In addition, do you think it's necessary to squash the commits if they can be merged?

eknabevcc:
Haha, I see your message now. 
Good that we agreed :)
Yes, we could have squashed, but that's no problem. We forget about it sometimes so esmini history is not perfect at all anyway.

PR #297, opened 2022-07-06
Fix error messages to show junction id.

mljack:
Fix typos in error msgs.


eknabevcc:
Yes, thanks a lot for finding and fix!

PR #323, opened 2022-09-20
Added wheel_angle and wheel_rotation to SE_ScenarioObjectState

### LINKED ISSUE #322: Get wheel angle via esminiLib.hpp.
I'd like to retrieve the `ObjectInfoStruct::wheel_angle` through the c++ interface.

I'm willing to take over the implemenation part myself.

For me the best fitting place would be to enlarge the `SE_ScenarioObjectState` struct and set it inside `copyStateFromScenarioGateway` in esminiLib.cpp, but this will not only break the public API but also might cause some very ugly side effects when the user of the lib overrides a larger memory area than previously.

mickmack1213:
Closes #322 

@eknabe TODO: Update the esmini.github.io Python Binding section. The ctypes representation of SE_ScenarioObjectState [here](https://esmini.github.io/#_get_object_states)


eknabevcc:
Thanks for contribution! 

Included in release [v2.26.5](https://github.com/esmini/esmini/releases/tag/v2.26.5).

Code example [get_states.py](https://github.com/esmini/esmini/blob/master/EnvironmentSimulator/code-examples/hello_world/get_states.py) (and [User guide](https://esmini.github.io/#_get_object_states)) updated.

PR #336, opened 2022-10-20
Free memory leak

csawp:


eknabevcc:
Thanks for the contribution! 

A minor fix in a few unit tests is needed to avoid double free ops which cause build error on Linux/gcc. That's taken care of by commit 645d337550aad754b4860872ada1e4268caaee79 on master branch. 

So, even if CI check fails on this branch it should be resolved when rebased to new master.

Issue #311, opened 2022-07-29
Add SE_RegisterStoryBoardElementStateChangeCallback to esminiLib


mickmack1213:
I'm currently working on a project where I want to simulate and monitor many OpenSCENARIO Files automatically via a python-script.

For this a better way to know when the simulation has finished would be great. Currently I'm tracking if any events were started and did not finish, but that's not really great...

So if I could monitor the States of the StoryboardElements would be pretty nice.

I'd propose a callback signature something like:
`void (*fnPtr)(const char* name, ElementType type, State state)`

And I would place the call to the callback into `StoryBoardElement::SetState`.

What do you think of this idea? Would you be willing to change the public API? If yes I can make a push request and implement it.

eknabevcc:
This sounds like a good idea. However, to my understanding it is similar to the already existing callback: [SE_RegisterEventCallback](https://github.com/esmini/esmini/blob/8be51cc09fd44af3dcef3e3dcc7dbffa18dfbca5/EnvironmentSimulator/Libraries/esminiLib/esminiLib.hpp#L849) - which was kindly added by @cfschaller in PR #249.

To my understanding and in spite of its name it handles all types of StoryBoardElement types, not only events. It reports back timestamp and a flag indicating whether the scenario element was started or ended.

Two difference I can see are: 1. Your proposal explicitly reports the element type. 2. The current implementation includes timestamp (instead).
Another difference is where (and in effect when) the callback happens.

Before implementing something similar or overlapping, can you please review that existing functionality to see if it covers your need or not?

mickmack1213:
Hmm, yeah it is similar, but I think I have to disagree with you, the current implementation does not provide this capability :thinking: 

At least if I scan for usages of the callback in CLion it only shows the following and all other matches with plain text search for `eventCallback` don't seem to do this also.

![grafik](https://user-images.githubusercontent.com/20105925/183418709-8f626dfc-8819-4151-b6ce-3218b671bb76.png)

It is true that this will work for all classes inheriting from `Event`, but for many StoryBoardElement types that's not the case. And I don't think it's good to change that.

My experience, if not really great but I did a few tests with the python interface, also showed, that those callbacks are only triggered for events. I put a log below, the output of my EventCallback is marked with `####` and see how no event is triggered for any state change of `MySequence` or `MyAct`

```
0.100: Trigger /------------------------------------------------
0.100: no name SimulationTimeCondition == true, 0.1000 > 0.00 edge: rising
[...]

PR #344, opened 2022-11-04
Removed SE_RegisterEventCallback, replaced the respective test 

cfschaller:
Hi Emil, 
As discussed in #311, here is the PR to remove the old callback (SE_RegisterEventCallback).

Whilst removing the respective test I saw that there is another test for SE_RegisterStoryBoardElementStateChangeCallback, but none for SE_RegisterConditionCallback. Therefore, instead of removing it, I reused the existing test for SE_RegisterEventCallback to create one for SE_RegisterConditionCallback.

Feel free to make further adjustments or remove the test, if you do not see it as useful.

Thanks again for your great work,
Christoph


eknabevcc:
Perfect. Thanks for fixing this, and also adding the missing test case for condition callback.

PR #354, opened 2022-11-21
feat: modify vehicle maxDeceleration set err

bkcarlos:
next read maxDeceleration,so judge  maxDeceleration


eknabevcc:
Excellent find! Thanks for the fix 👍

PR #375, opened 2023-01-17
fix typos in user_guide.adoc

a120471:
As I was reading the user guide, I found several typos based on my understanding, please take a look whether my changes are correct. Thanks.

### DIFF
--- a/docs/user_guide.adoc
+++ b/docs/user_guide.adoc
@@ -574,7 +574,7 @@ Only trace file way is described here. To save OSI data, add argument `--osi_fil
 
 ``./bin/esmini --window 60 60 800 400 --osc ./resources/xosc/cut-in.xosc --osi_file``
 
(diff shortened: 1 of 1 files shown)


eknabevcc (commented):
Thanks a lot for contributing these fixes!

All except the following one are sure corrections and improvements: 
I think SpeedProfileEntry time 0.0 is correct. It’s representing a relative time, and since the complete action is triggered at t=2.0 the resulting time for first entry becomes 2.0 + 0.0.
Also it’s aligned with the scenariofile:
https://github.com/esmini/esmini/blob/master/resources/xosc/speed-profile.xosc

If you agree, can you revert just that change and I’ll merge the PR 👍

a120471:
Thanks for your explanation reply, I've reverted that wrong change back.

eknabevcc:
Excellent. Will merge now. Published User guide will be updated accordingly at latest with next release.

PR #392, opened 2023-02-13
Fix Timeout Misinterpretation in Unix

raaphy:
The current implementation of UDPServer::UDPServer in Unix systems sets _tv.usec_ equal to _timeoutMs_, which is intended to be in milliseconds. However, _timeval.usec_ is expecting the value to be in microseconds.

This pull request fixes this issue by converting timeoutMs to microseconds before setting it to tv.usec. Additionally, it now supports timeouts greater than 1 second.


eknabevcc:
I hope you will make another attempt! I just reviewed your pull request, and it looked great. 

Unfortunately something went wrong on Windows build. 

If you make another PR, or re-open this one we can try solve it together.

Maybe you can try rebase on master instead of merge, but I'm not sure that will help.

eknabevcc:
Ah, now I see it (the next one) :) 👍

PR #393, opened 2023-02-13
Fix incorrect use of timeout parameter in unix.

raaphy:
The current implementation of UDPServer::UDPServer in Unix systems sets tv.usec equal to timeoutMs, which is intended to be in milliseconds. However, timeval.usec is expecting the value to be in microseconds.

This pull request fixes this issue by converting timeoutMs to microseconds before setting it to tv.usec. Additionally, it now supports timeouts greater than 1 second.


eknabevcc:
I think the hick-up root cause is that the PR is one commit behind. Please leave this as is, and I'll try to merge it manually (command line). Because then we can rebase on master.

It's already in the pipe: https://github.com/esmini/esmini/commits/fix_socket_timeout

eknabevcc:
Th PR has been merged manually on master (e5e19acfedd83563c276a84bf0fc655100ed456f)

The solution for the hick-up on Windows build was to rebase the branch on master.

Thanks for nice contribution!

PR #420, opened 2023-04-11
Add OSI model_reference

### LINKED ISSUE #419: OSI model_reference missing in OSI Ground Truth
The model_reference in OSI gives the path to a corresponding 3D model file for moving and stationary objects as well as the stationary environment. This is currently missing in the OSI API.

ClemensLinnhoff:
- [x] Added model reference to ground truth moving objects
- [x] Added model reference for stationary environment

This PR closes #419.


eknabevcc:
The root cause for CI fail is changes to the dat file format. I suggest that we keep 3dmodel filename out of the dat file in order to keep it slim. We use model_id instead to identify what model to use for replaying the scenario.

Another minor thing is changed content in OSI data which affects the size of the OSI files, hence size check fail.

I [fixed](https://github.com/esmini/esmini/commits/feature/pr_420_mod) these to topics in branch: [pr_420_mod](https://github.com/esmini/esmini/tree/feature/pr_420_mod)

It [passes the CI](https://github.com/esmini/esmini/actions/runs/4698539060) now.

You can decide whether you want to apply same changes on your PR, or if we should merge "my" branch instead. (similar situation as for #422)

ClemensLinnhoff:
Thank you! We can merge your branch.

eknabevcc:
Merged by commit a7220d06cc6bf1ea4e4dca660144f2e97843e938
Close PR

PR #428, opened 2023-05-02
Added OpenMSL to Related Work

ClemensLinnhoff:
I added OpenMSL to the Related Work section in the Readme. I think this will benefit the users as they can see, how esmini is used in a co-simulation with other FMUs directly in a GitHub action to test sensor models.

Further applications e.g. für OpenDrive and OpenScenario files will follow in OpenMSL.

I also linked esmini in the [Related Work in OpenMSL](https://github.com/openMSL/.github/blob/main/doc/related_work.md).


eknabevcc:
Excellent! Thanks for updating related work on both sides 👍

It will be visible on esmini GitHub soon by next release (when dev is merged on master)

PR #429, opened 2023-05-08
fix: add includes to esminiJS CMakeLists.txt

HQidea:
Make building esminiJS work again

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,5 @@ EnvironmentSimulator/Libraries/esminiJS/build*/*
 support/generated
 pylint_report.txt
 CMakeUserPresets.json
+.DS_Store
+EnvironmentSimulator/Libraries/esminiJS/example/esmini.js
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -35,4 +35,6 @@ out/build/**
 EnvironmentSimulator/Libraries/esminiJS/build*/*
 support/generated
 pylint_report.txt
-CMakeUserPresets.json
\ No newline at end of file
+CMakeUserPresets.json
```
eknabevcc, line 38: Regarding the three added files to the ignore list:
I googled .DS_Store and now understand that it's a Mac system file, which clearly can be ignored 👍
But the other two:
- CMakeUserPresets.json
- esmini.js
I would suggest they are not ignored. I.e. they should be removed from the list. Would that be OK with you? Or do you have some reasons why they should be ignored?

eknabevcc:
Thanks for the fixes!

I added one review comment on the .gitignore changes. Please check my comment.

I just realized that the Mac OS version we specified for GitHub actions is not supported anymore, and hence needs to be updated (from 10.5 to 11.0) - that's the reason why the checks did not fully succeeded. But that's no problem, I will approve anyway, as soon as when we sorted out the .gitignore question.

.gitignore
eknabevcc: Thanks for clarifications. Indeed CMakeUserPresets.json was already included. My mistake. 
Same for esmini.js - you're completely right, generated files should be ignored.

PR #466, opened 2023-08-19
[roadmngr] const correctness on class methods

brifsttar:
Hey Emil,

A while ago I mentioned that I like (and use) [const correctness](https://isocpp.org/wiki/faq/const-correctness) a lot, which over time lead me to add it here and there around RoadManager when I needed it. But since I don't do that very often and nearly always forget to cherry-pick and make a PR out of it, well...

So I figured I might as well go over all of RoadManager and do it once and for all. Some notes:
* I only did it for method signatures, *not* for method parameters
* I skipped over some classes I didn't know of (I'm out of date on the standard it seems 😅 ), like `Shape` and `OutlineCorner`
* Interestingly, some methods that looked like they could be `const`'d actually couldn't, like `Position::GetNumberOfRoadsOverlapping()` ([link](https://github.com/esmini/esmini/blob/41f5f0834b719b7444c4cf1263c6a9d2f2c01044/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp#L9057))

Thanks again for all your work. RoadManager is at the core of what we do in Unreal, and I can't imagine what we'd do without it. The library has also proven very robust, even with the more and more complex road networks we build, so it's a real testament to the quality of your work.

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -5078,7 +5078,8 @@ int OpenDrive::GetTrackIdByIdx(int idx) const
     return 0;
 }
 
-bool OpenDrive::IsIndirectlyConnected(int road1_id, int road2_id, int*& connecting_road_id, int*& connecting_lane_id, int lane1_id, int lane2_id) const
+bool OpenDrive::IsIndirectlyConnected(int road1_id, int road2_id, int*& connecting_road_id, int*& connecting_lane_id, int lane1_id, int lane2_id)
+    const
 {
     Road*     road1 = GetRoadById(road1_id);
     Road*     road2 = GetRoadById(road2_id);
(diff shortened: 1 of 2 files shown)


eknabevcc:
Great! Thanks for this nice code improvement!
Good to hear roadmanager works ok. I think you and other users has contributed by using and pushing it, finding bugs and limitations along the way :)

eknabevcc:
BTW, regarding Position::GetNumberOfRoadsOverlapping(). It actually writes some info to the object about any overlapping roads, so that information about the overlapping roads can be retrieved in a succeeding step. 

See code example [rm-basic.cpp](https://github.com/esmini/esmini/blob/41f5f0834b719b7444c4cf1263c6a9d2f2c01044/EnvironmentSimulator/code-examples/rm-basic/rm-basic.cpp#L107)

Of course, this could be refactorized so that all information is returned in one call, but we try to stick with primitive types in the library APIs - to maximize portability. And I don't see a simple solution for dynamic (size not known on beforehand) lists/arrays.

Issue #475, opened 2023-09-21
OSMP_FMU: osiversion will not be replaced


localhorst87:
In CMakeLists.txt there is no directive to replace @OSIVERSION@, so in the modelDescription.xml, the osi-version attribute in the osmp element will be empty for each variable.

```
<Tool name="net.pmsf.osmp" xmlns:osmp="http://xsd.pmsf.net/OSISensorModelPackaging">
   <osmp:osmp version="1.3.0" osi-version=""/>
</Tool>
``` 

As some tools like AVL Model.CONNECT that already implemented OSI require this osi-version attribute to interpret the variables `.base.lo` + `.base.hi` + `.size` as a common binary input, this should be fixed (see the image).

![image](https://github.com/esmini/esmini/assets/43797690/ad4c7d51-feca-4ad3-a40e-723a0239f522)

@ClemensLinnhoff

eknabevcc:
I guess it's it should be specified in the [OSMP_FMU/CMakeLists.txt](https://github.com/esmini/esmini/blob/dev/OSMP_FMU/CMakeLists.txt)?

Could you suggest solution, perhaps even in a PR?

ClemensLinnhoff:
@eknabevcc we typically include OSI as a submodule in our simulation models (we talked about this some while back). When it is a submodules we can access the [version file](https://github.com/OpenSimulationInterface/open-simulation-interface/blob/master/VERSION) directly from the submodule and set it as a cmake variable in the cmakelists. (See [this example](https://github.com/openMSL/sl-1-0-sensor-model-repository-template/blob/main/CMakeLists.txt#L10)) That variable is then used to set it in the modelDescription. Since in esmini OSI is included as a binary, it does not work that way. Do you have a suggestion how to automatically get the information, which OSI version is currently used in esmini?

PR #476, opened 2023-09-25
Fix versioning in modelDescription

ClemensLinnhoff:
In an OSMP modelDescription, there are 3 different version to set. None of them are set in the current master branch, as pointed out by https://github.com/esmini/esmini/issues/475.

In this PR 3 versions are set in the CMakeLists of the OSMP FMU:
- VERSION: Version of the FMU itself. We could set it to the same version number as esmini itself. @eknabe can we access this directly in the repository? the version.txt does not contain the actual version number.
- OSIVERSION: OSI version used by esmini. As already mentioned in the corresponding issue: @eknabe any idea on how to set this automatically?
- OSMPVERSION: Can be set manually, depending on which OSMP version is actually used, as implemented in this PR,

### DIFF
--- a/OSMP_FMU/CMakeLists.txt
+++ b/OSMP_FMU/CMakeLists.txt
@@ -3,6 +3,8 @@ cmake_minimum_required(VERSION 3.5)
 set(TARGET EsminiOsiSource)
 project(${TARGET})
 
+set(VERSION "2.0.0")
+
 set(STATIC_LINKING TRUE)
 
 if(LINUX)
@@ -79,6 +81,9 @@ set(VERBOSE_FMI_LOGGING
 set(DEBUG_BREAKS
     OFF
     CACHE BOOL "Enable debugger traps for debug builds of FMU")
+
+set(OSIVERSION "3.5.0")
+set(OSMPVERSION "1.4.0")
 
 string(TIMESTAMP FMUTIMESTAMP UTC)
 string(MD5 FMUGUID modelDescription.in.xml)
--- a/OSMP_FMU/CMakeLists.txt
+++ b/OSMP_FMU/CMakeLists.txt
@@ -3,7 +3,9 @@ cmake_minimum_required(VERSION 3.5)
 set(TARGET EsminiOsiSource)
 project(${TARGET})
 
-set(VERSION "2.0.0")
+set(VERSION "2.32.1")
+set(OSIVERSION "3.5.0")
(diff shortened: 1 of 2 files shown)


ClemensLinnhoff:
I currently hardcoded the version numbers for all three, just to show that it generally works. They should now all be contained in the modelDescription. @localhorst87 can you verify this?

@eknabe could you assist on setting the esmini version and the osi version automatically in the CMakeLists?

ClemensLinnhoff:
I currently hardcoded the version numbers for all three, just to show that it generally works. They should now all be contained in the modelDescription. @localhorst87 can you verify this?

@eknabe could you assist on setting the esmini version and the osi version automatically in the CMakeLists?

eknabevcc:
version.txt should contain tag. Example:

```
ESMINI_GIT_REV="e56d4c1"
ESMINI_GIT_TAG="v2.31.1"
ESMINI_GIT_BRANCH="tags/v2.31.1"
ESMINI_BUILD_VERSION="3246"
```
But if the current commit (HEAD) is not on a tag it will be empty.

Regarding OSI version, I'm not sure how to grab version nr easily. 

I'm preparing a release for today. I suggest we merge what you have and we can refine the versioning later. What do you think?

ClemensLinnhoff:
> I'm preparing a release for today. I suggest we merge what you have and we can refine the versioning later. What do you think?

What will be the Release Tag? 2.32.1? Then I can put this already in the CMakeLists. And for this release it will match.

ClemensLinnhoff:
And are you still on OSI 3.5 or will you update to 3.6?

eknabevcc:
> And are you still on OSI 3.5 or will you update to 3.6?

Still [v3.5.0](https://github.com/esmini/esmini/blob/6c252273c5aa2422d5629a59febeee436fa74ee2/scripts/generate_osi_libs.sh#L31C1-L31C18)

eknabevcc:
> > I'm preparing a release for today. I suggest we merge what you have and we can refine the versioning later. What do you think?
> 
> What will be the Release Tag? 2.32.1? Then I can put this already in the CMakeLists. And for this release it will match.

Ah, yes, release tag will be v2.32.1

ClemensLinnhoff:
I changed to hardcoded versions. Now it fit to the next release and can be merged.

eknabevcc:
I can add a dynamic version setting, however I can't edit files in this branch/repo. I'll do it after merge.

PR #528, opened 2024-01-31
Feature/osi traffic cmd osmp

ClemensLinnhoff:
Add OSI TrafficCommand to OSI API as well as the OSMP FMU.


eknabevcc:
Great! Thanks for quick updates.

Actually I already merged the OSITrafficCmd stuff on dev. So I wonder what will happen if I now merge this complete PR including same commit... (I see the commit hash differ, so it could get crazy I'm not sure)

Would it be possible for you to update the PR so that only your (the 2:nd) commit remains?

ClemensLinnhoff:
I am not sure, how I can remove commits in-between. I just merged dev back into this branch. Shouldn't it work this way? The diff only shows the changed from the OSMP stuff I added.

eknabevcc:
You're right. I also double-checked by rebasing the branch on dev locally. 
So let's go! It's merged 👍

PR #537, opened 2024-02-26
Fix CommonMini version generation

GavriloviciEduard:
Build fails sometimes because GIT_REV contains trailing spaces


eknabevcc:
Thanks for this fix! 

It's included in tag/release [v2.37.0](https://github.com/esmini/esmini/releases/tag/v2.37.0).

PR #575, opened 2024-05-30
Updated swarm traffic to change heading based on RHT or LHT

joe-mitchell227:
We had an issue where the swarm traffic action was only working for RHT. I have added a check to find the traffic direction from ODR and swapped the heading around for LHT. 

After visualising this for scenarios with LHT and RHT it then worked. I'm not sure if there is anything else that would need changing for completeness apart from the SetHeadingRelativeRoadDirection?

I have also left it to default to RHT when the traffic direction is undefined, you may want to change this.


eknabevcc:
Thanks for this fix!

PR #583, opened 2024-06-27
Add arguments to ignore z, pitch and roll inputs from OSC

lrstttl:
We have the use case where an input OpenScenario contains z, pitch and roll values which do not perfectly fit to the underlying OpenDrive, i.e., the vehicle is flying or its orientation does not match the road.

Instead of working on the OpenScenario, I added arguments `ignore_z`, `ignore_p`, `ignore_r` to ignore these values and make use of esmini's relative positioning on the road.

Usage is not affected if the arguments are not provided.

Running a scenario which has "wrong" z inputs:
```
esmini --osc input_z+5.xosc --window 600 600 600 400
```
image

Running it and ignoring the "wrong" z inputs:
```
esmini --osc input_z+5.xosc --window 600 600 600 400 --ignore_z
```
image

**However, I still notice a bug where the connected tow and trailer vehicles already had this feature, i.e. they are currently always placed relative to the road and ignore the z from the towing vehicle**

### DIFF
--- a/EnvironmentSimulator/Libraries/esminiLib/esminiLib.cpp
+++ b/EnvironmentSimulator/Libraries/esminiLib/esminiLib.cpp
@@ -537,7 +537,7 @@ extern "C"
         return InitScenario();
     }
 
-    static int AddCommonArguments(int disable_ctrls, int use_viewer, int threads, int record)
+    static int AddCommonArguments(int disable_ctrls, int use_viewer, int threads, int record, bool ignore_z, bool ignore_p, bool ignore_r)
     {
         (void)record;
         if ((use_viewer & 1) == 0)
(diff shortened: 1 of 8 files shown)


eknabevcc:
I like the idea! 

Implementation wise I'd prefer another way of setting the flags instead of extra arguments for SE_Init(). I will check for an alternative approach. 

Meanwhile, can you elaborate slightly on the trailer related bug?

eknabevcc:
I don't know how to commit to your PR. But I reworked slightly in a new branch:

[feature/pr_583](https://github.com/esmini/esmini/commits/feature/pr_583/)

The idea is to make use of the Options class to keep the values. The options can be specified through API prior to calling SE_Init() like:

`SE_SetOption("ignore_z");`

It would be great if you could try the branch and see if it still works as expected.

(this new solution can actually be reused for other options as well, if it works without side effects...)

eknabevcc:
The implementation in the branch mentioned above has been added to master and is included in release [v2.37.15](https://github.com/esmini/esmini/releases/tag/v2.37.15). 

Hopefully this first shot works good enough. Please test and let us know if it works as expected or not.

PR #589, opened 2024-07-08
Fix the issue of transition dynamics not work in relative speed action when the act to which it belongs is before the act to which its rerence entity belongs

wubinwab:
Hello,
    I’ve encountered a problem where, if the RelativeTargetSpeed’s reference object is set after the current object, even if TransitionDynamics are set, RelativeTargetSpeed does not use the acceleration specified in the dynamics when adjusting speed, but instead directly uses the current speed of the target. Here is a scenario to reproduce the issue:
```
<?xml version='1.0' encoding='utf-8'?>
<OpenSCENARIO xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="OpenScenario.xsd">
    <FileHeader author="author" date="2024-07-08T11:51:07.575696" description="scenario_description" revMajor="1" revMinor="2" />
    <ParameterDeclarations>
        <ParameterDeclaration name="ego_lateral_state" parameterType="int" value="0" />
        <ParameterDeclaration name="ego_urgency_level" parameterType="int" value="0" />
    </ParameterDeclarations>
    <CatalogLocations>
        <VehicleCatalog>
            <Directory path="../xosc/Catalogs/Vehicles" />
        </VehicleCatalog>
        <MiscObjectCatalog>
            <Directory path="../xosc/Catalogs/MiscObjects" />
        </MiscObjectCatalog>
        <ControllerCatalog>
            <Directory path="../xosc/Catalogs/Controllers" />
        </ControllerCatalog>
        <PedestrianCatalog>
            <Directory path="../xosc/Catalogs/Pedestrians" />
        </PedestrianCatalog>
    </CatalogLocations>
    <RoadNetwork>
        <LogicFile filepath="../resources/xodr/straight_500m.xodr" />
    </RoadNetwork>
    <Entities>
        <ScenarioObject name="ego">
            <CatalogReference catalogName="VehicleCatalog" entryName="semi_truck" />
        </ScenarioObject>
        <ScenarioObject name="actor_1">
            <CatalogReference catalogName="VehicleCatalog" entryName="car_red" />
        </ScenarioObject>
    </Entities>
    <Storyboard>
        <Init>
            <Actions>
                <Private entityRef="ego">
                    <PrivateAction>
                        <TeleportAction>
                            <Position>
                                <LanePosition laneId="-1" offset="-0.03310905615197002" roadId="1" s="20.785420001895005">
                                    <Orientation h="0.0" type="relative" />
                                </LanePosition>
                            </Position>
                        </TeleportAction>
                    </PrivateAction>
                    <PrivateAction>
                        <LongitudinalAction>
                            <SpeedAction>
                                <SpeedActionDynamics dynamicsDimension="time" dynamicsShape="step" value="0.0" />
                                <SpeedActionTarget>
                                    <AbsoluteTargetSpeed value="22.22222222222222" />
                                </SpeedActionTarget>
                            </SpeedAction>
                        </LongitudinalAction>
                    </PrivateAction>
                </Private>
                <Private entityRef="actor_1">
                    <PrivateAction>
                        <TeleportAction>
                            <Position>
                                <LanePosition laneId="-1" offset="0.10324825307863628" roadId="1" s="52.542815709409226">
                                    <Orientation h="0.0" type="relative" />
                                </LanePosition>
                            </Position>
                        </TeleportAction>
                    </PrivateAction>
                    <PrivateAction>
                        <LongitudinalAction>
                            <SpeedAction>
                                <SpeedActionDynamics dynamicsDimension="time" dynamicsShape="step" value="0.0" />
                                <SpeedActionTarget>
                                    <AbsoluteTargetSpeed value="22.22222222222222" />
                                </SpeedActionTarget>
                            </SpeedAction>
                        </LongitudinalAction>
                    </PrivateAction>
                </Private>
            </Actions>
        </Init>
        <Story name="story1">
            <Act name="act8">
                <ManeuverGroup maximumExecutionCount="1" name="maneuver_group6">
                        <EntityRef entityRef="ego" />
                    </Actors>
                    <Maneuver name="maneuver5">
                        <Event maximumExecutionCount="1" name="event3" priority="parallel">
                            <Action name="action4_relative_speed">
                                <PrivateAction>
                                    <LongitudinalAction>
                                        <SpeedAction>
                                            <SpeedActionDynamics dynamicsDimension="rate" dynamicsShape="linear" value="0.02" />
                                            <SpeedActionTarget>
                                                <RelativeTargetSpeed continuous="true" entityRef="actor_1" speedTargetValueType="delta" value="0.0" />
                                            </SpeedActionTarget>
                                        </SpeedAction>
                                    </LongitudinalAction>
                                </PrivateAction>
                            </Action>
                            <StartTrigger>
                                <ConditionGroup>
                                    <Condition conditionEdge="none" delay="0.0" name="trigger2">
                                        <ByValueCondition>
                                            <SimulationTimeCondition rule="greaterThan" value="0.1" />
                                        </ByValueCondition>
                                    </Condition>
                                </ConditionGroup>
                            </StartTrigger>
                        </Event>
                    </Maneuver>
                </ManeuverGroup>
                <StartTrigger>
                    <ConditionGroup>
                        <Condition conditionEdge="rising" delay="0.0" name="start_trigger7">
                            <ByValueCondition>
                                <SimulationTimeCondition rule="greaterThan" value="0.0" />
                            </ByValueCondition>
                        </Condition>
                    </ConditionGroup>
                </StartTrigger>
                <StopTrigger />
            </Act>
            <Act name="act14">
                <ManeuverGroup maximumExecutionCount="1" name="maneuver_group12">
                        <EntityRef entityRef="actor_1" />
                    </Actors>
                    <Maneuver name="maneuver11">
                        <Event maximumExecutionCount="1" name="event9" priority="parallel">
                            <Action name="action10_accel">
                                <PrivateAction>
                                    <LongitudinalAction>
                                        <SpeedAction>
                                            <SpeedActionDynamics dynamicsDimension="rate" dynamicsShape="linear" value="2.0" />
                                            <SpeedActionTarget>
                                                <AbsoluteTargetSpeed value="25.0" />
                                            </SpeedActionTarget>
                                        </SpeedAction>
                                    </LongitudinalAction>
                                </PrivateAction>
                            </Action>
                            <StartTrigger>
                                <ConditionGroup>
                                    <Condition conditionEdge="none" delay="0.0" name="trigger2">
                                        <ByValueCondition>
                                            <SimulationTimeCondition rule="greaterThan" value="0.1" />
                                        </ByValueCondition>
                                    </Condition>
                                </ConditionGroup>
                            </StartTrigger>
                        </Event>
                    </Maneuver>
                </ManeuverGroup>
                <StartTrigger>
                    <ConditionGroup>
                        <Condition conditionEdge="rising" delay="0.0" name="start_trigger13">
                            <ByValueCondition>
                                <SimulationTimeCondition rule="greaterThan" value="0.0" />
                            </ByValueCondition>
                        </Condition>
                    </ConditionGroup>
                </StartTrigger>
                <StopTrigger />
            </Act>
            <Act name="act21">
                <ManeuverGroup maximumExecutionCount="1" name="maneuver_group19">
                        <EntityRef entityRef="actor_1" />
                    </Actors>
                    <Maneuver name="maneuver18">
                        <Event maximumExecutionCount="1" name="event16" priority="parallel">
                            <Action name="action17_accel">
                                <PrivateAction>
                                    <LongitudinalAction>
                                        <SpeedAction>
                                            <SpeedActionDynamics dynamicsDimension="rate" dynamicsShape="linear" value="1.0" />
                                            <SpeedActionTarget>
                                                <AbsoluteTargetSpeed value="13.88888888888889" />
                                            </SpeedActionTarget>
                                        </SpeedAction>
                                    </LongitudinalAction>
                                </PrivateAction>
                            </Action>
                            <StartTrigger>
                                <ConditionGroup>
                                    <Condition conditionEdge="none" delay="0.0" name="trigger15">
                                        <ByValueCondition>
                                            <StoryboardElementStateCondition state="endTransition" storyboardElementRef="action10_accel" storyboardElementType="action" />
                                        </ByValueCondition>
                                    </Condition>
                                </ConditionGroup>
                            </StartTrigger>
                        </Event>
                    </Maneuver>
                </ManeuverGroup>
                <StartTrigger>
                    <ConditionGroup>
                        <Condition conditionEdge="rising" delay="0.0" name="start_trigger20">
                            <ByValueCondition>
                                <SimulationTimeCondition rule="greaterThan" value="0.0" />
                            </ByValueCondition>
                        </Condition>
                    </ConditionGroup>
                </StartTrigger>
                <StopTrigger />
            </Act>
        </Story>
        <StopTrigger>
            <ConditionGroup>
                <Condition conditionEdge="rising" delay="0.0" name="stop_simulation">
                    <ByValueCondition>
                        <SimulationTimeCondition rule="greaterThan" value="60.0" />
                    </ByValueCondition>
                </Condition>
            </ConditionGroup>
        </StopTrigger>
    </Storyboard>
    <VariableDeclarations />
</OpenSCENARIO>
```

### DIFF
--- a/EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCPrivateAction.cpp
+++ b/EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCPrivateAction.cpp
@@ -1097,6 +1097,11 @@ void LongSpeedAction::Step(double simTime, double dt)
         }
     }
 
+    if (target_->type_ == Target::TargetType::RELATIVE_SPEED)
+    {
+        target_speed_reached_ = false;
+    }
+
     object_->SetSpeed(ABS_LIMIT(new_speed, object_->performance_.maxSpeed));
 
     if (target_speed_reached_ &&


eknabevcc:
Thanks, seems like an interesting find! 
I'll look into it as soon as possible.

eknabevcc:
Thanks for the provided scenario. It really helps investigation.

To clarify interpretation of the relative SpeedAction: 

esmini will only apply transition once, while reaching the target speed. Once reached the target speed esmini will either 1. end the action (continuous="false") or 2. exactly maintain the speed of referenced entity (continuous="true").

While investigating we tried a slightly different approach: Run SpeedAction without continuous mode and instead restart the SpeedAction as soon as the speed deviates again. I.e. looping:
1. Reach speed of referenced entity
2. End action and stay on this speed until speed difference is again significant

However, we found two issues in esmini:
1. Target speed not reset when action is repeated
2. Target speed not continuously updated while reaching relative target speed (I think this is the issue you encountered)

These issues are now fixed by commits ee1cc5e6924d63a9ec461f1a4614b558037ba348 and 40cbc7e8eb8962962f124314e83a6d277d4bdc17, both merged on dev branch (will be part of next release). 

Updated behavior: Reference for the relative speed will always be the current speed of referenced entity.

The scenario [repeated_relative_speed.xosc](https://github.com/esmini/esmini/blob/dev/EnvironmentSimulator/Unittest/xosc/repeated_relative_speed.xosc) is based on yours but modified according to the above strategy. To demonstrate the scheme, the target car (actor_1) is aggressively accelerating and decelerating three times. Ego will try to follow the speed changes. It's now being used in a smoke test case.  Note that this scenario will work only with the applied commits.

Summary: To my understanding the commits solves the root cause of the problem you encountered, but in a slight different way than your proposed code change (cleaning up code instead of adding code to fix the issue caused by old code).

Please check whether your scenario works as intended with the fixes. Alternatively you can maybe make use of the approach in the test scenario. And again, thanks for reporting the issue!

wubinwab:
Thank you for your reply. I have tested with your submissions and found that using your scenario is feasible. However, using the scenario submitted in my issue still does not work as the rear vehicle did not maintain relative speed with the front vehicle.
Let me first introduce the expected behavior of my scenario: I expect the rear vehicle to maintain a relative speed with the front vehicle at **the acceleration** I set. The problem I encountered before was that although the rear vehicle maintained a relative speed with the front car, **it did not use the set acceleration, but directly used the speed of the front vehicle.**
I compared your demo scenario with mine and found that if the start trigger for the relative speed action is changed to a time trigger like mine, the action of maintaining relative speed will not occur.
Can you double check the issue or check if my submission is feasible? Thank you！

eknabevcc:
Indeed! Thanks for testing and reporting failure. You encountered a bug introduced in last commits that prevented continuous mode altogether. So that's why the action in your scenario is terminated as soon as the target relative speed is reached. 

The bug is fixed by commit 205a66bc2b6c7d8e3b52d7ac149d58c88ad13fca.

Regarding behavior of the action: As previously stated, the transition only applies initially. Once relative speed is reached, the relative speed will be maintained.

See plot below of modified variant of your scenario, [i589_mod2.zip](https://github.com/user-attachments/files/16189151/i589_mod2.zip). I changed 1. ego initial speed to 21.222, 1 m/s slower than actor_1 22.2222 and rate from 0.02 to 5.0. So initially ego will accelerate with rate 5.0 m/s^2. Once reaching the relative speed 0 (as specified) transition is done and it will just maintain relative speed. In other words: Vehicle will accelerate or decelerate until it reach specified relative speed, then maintain it. 


plot

The proposed change I agree is reasonable when transition dynamicsDimension is "rate". However it does not harmonize with dynamicDimension "distance" or "time", in which cases you expect an end of the transition phase. Potentially the continuous acceleration mode could be applied for "rate" only. But that adds some complexity that also can cause confusion.

wubinwab:
Thank you for your quick reply. So I need to cherry-pick the three commits to do verification?

eknabevcc:
To test with all commits applied, just pull latest from dev branch.

Regarding acceleration, I think I already outlined my thinking above. Summary: That it (“rate”) belongs to the transition, which is not always intended to go on forever.

PR #604, opened 2024-08-07
Update esmini wrapper

KonstantinBlenzTT:
We already use this functionality offered by the esminiLib and now just need a updated wrapper.


eknabevcc:
Looks good, thanks for contributing

PR #607, opened 2024-08-14
Take object orientation into account in OSI moving object position

### LINKED ISSUE #568: Correctly take orientation into account for OSI output
As far as I saw in the code and from my own experiments, it seems that for the calculation of the OSI output, esmini assumes that the car is oriented perfectly horizontal. Any non-zero pitch and roll result in wrong positions in OSI.

OSI references all object positions by the location of their bounding box center. As esmini uses OpenSCENARIO coordinates (if I'm not mistaken) for all objects, which use the center of the rear axle projected orthogonally onto the ground plane of the object as the reference point, a coordinate transformation is needed to get the bounding box centers of the objects for OSI. This is done by taking the bounding box center offset, rotating it into the frame of the object and adding it to the OpenSCENARIO reference point.

Currently, a simplified version of exactly this procedure is done in `OSIReporter::UpdateOSIMovingObject` (**OSIReporter.cpp:860**):
```cpp
// Set OSI Moving Object Position
obj_osi_internal.mobj->mutable_base()->mutable_position()->set_x(
        objectState->state_.pos.GetX() + static_cast<double>(objectState->state_.info.boundingbox.center_.x_) * cos(objectState->state_.pos.GetH()));
obj_osi_internal.mobj->mutable_base()->mutable_position()->set_y(
        objectState->state_.pos.GetY() + static_cast<double>(objectState->state_.info.boundingbox.center_.x_) * sin(objectState->state_.pos.GetH()));
obj_osi_internal.mobj->mutable_base()->mutable_position()->set_z(
        objectState->state_.pos.GetZ() + static_cast<double>(objectState->state_.info.boundingbox.dimensions_.height_) / 2.0);
```

However, only the heading is taken into account and pitch and yaw are assumed to be zero. This is fine or at least approximately ok for most cases, when the road is completely flat. Nevertheless, as soon as there are not negligible inclinations in the road, this produces erroneous results.

The correct implementation would take the offset vector, perform the full 3D rotation and add that result to the objectState position.

ClemensLinnhoff:
This PR is fixes #568.

I added a proper 3D coordinate transformation to transform from the esmini vehicle coordinate frame to the OSI coordinate frame. Now the full 3D orientation of the object is taken into account.

@d-hir could you please verify my implementation?

### DIFF
--- a/EnvironmentSimulator/Modules/ScenarioEngine/SourceFiles/OSIReporter.cpp
+++ b/EnvironmentSimulator/Modules/ScenarioEngine/SourceFiles/OSIReporter.cpp
@@ -858,12 +858,38 @@ int OSIReporter::UpdateOSIMovingObject(ObjectState *objectState)
     obj_osi_internal.mobj->mutable_base()->mutable_dimension()->set_length(objectState->state_.info.boundingbox.dimensions_.length_);
 
     // Set OSI Moving Object Position
-    obj_osi_internal.mobj->mutable_base()->mutable_position()->set_x(
-        objectState->state_.pos.GetX() + static_cast<double>(objectState->state_.info.boundingbox.center_.x_) * cos(objectState->state_.pos.GetH()));
-    obj_osi_internal.mobj->mutable_base()->mutable_position()->set_y(
-        objectState->state_.pos.GetY() + static_cast<double>(objectState->state_.info.boundingbox.center_.x_) * sin(objectState->state_.pos.GetH()));
-    obj_osi_internal.mobj->mutable_base()->mutable_position()->set_z(
-        objectState->state_.pos.GetZ() + static_cast<double>(objectState->state_.info.boundingbox.dimensions_.height_) / 2.0);
+    // As OSI defines the origin of the object coordinates in the center of the bounding box and esmini (as OpenSCENARIO)
+    // at the center of the rear axle, the position needs to be transformed.
+    // For the transformation the orientation of the object has to be taken into account.
+    std::array<std::array<double, 3>, 3> rotation_matrix{};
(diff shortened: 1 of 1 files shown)


ClemensLinnhoff:
@eknabe could you check why the macos-12 debug test fails? I don't understand the output since I did not make any changes to the files or lines that are marked.

eknabevcc:
Looks good! Thanks for this improvement.

I'd like to add a test case to check OSI pose, not only for regression but also to show the effect of the commit. In addition I'd like to reuse existing function for the actual transformation.

I noticed that I can't commit/edit in your branch. So instead I created a copy branch [feature/p607](https://github.com/esmini/esmini/tree/feature/p607). These are the steps I performed:
1. created the test case in dev branch
2. rebased p607 (the copy branch) to latest dev
3. squashed your commits into one
4. inserted test case updates (adapting to your changes) in your commit
5. added a commit which reuse the existing transform function - ensuring same result from previously added test

Commit history [here](https://github.com/esmini/esmini/commits/feature/p607/)

If you consent, I'll merge that p607 branch into dev.

ClemensLinnhoff:
Looks good. Though I think the changes to the readme and the doc probably do not belong to the feature?

eknabevcc:
True, those commits are already on [dev](https://github.com/esmini/esmini/commits/dev/), so they were added to the branch by the rebase.

eknabevcc:
Feature merged on dev branch from separate branch. See commits listed above.

d-hir:
(Original issue reporter here)

Sorry for the delay. I finally got around to testing this. From my quick tests (with the currently latest version 2.45.3) it looks like the OSI output is exactly (up to numerical precision errors) what I would expect, even on a road with quite high inclination. So thanks for fixing this!

PR #620, opened 2024-10-07
Use binary search to replace linear search to improve OpenDRIVE loading

mljack:
Optimize code speed for OpenDRIVE files with densely sampled in s.

`roadmanager::Position::LoadOpenDrive("road.xodr")`


eknabevcc:
Nice! Thanks for contribution. I had to educate myself on std::upper_bound function, lambda details and BinaryPredicate. But now I'm confident. I think your approach is a good example and can be applied in more similar cases, eventually.

PR #626, opened 2024-10-22
Fix lane offset

johschmitz:
See

https://publications.pages.asam.net/standards/ASAM_OpenDRIVE/ASAM_OpenDRIVE_Specification/latest/specification/11_lanes/11_04_lane_offset.html

and the documented known issue in the current release

https://esmini.github.io/#_reference_line_and_center_lane_while_using_laneoffset

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -1578,7 +1578,8 @@ int Road::GetLaneInfoByS(double s, int start_lane_section_idx, int start_lane_id
                         lane_info.lane_id_);
                 }
 
-                int new_lane_index = lane_section->GetClosestLaneIdx(s, t, 0, offset, true, laneTypeMask);
+                double lane_offset    = GetLaneOffset(s);
+                int    new_lane_index = lane_section->GetClosestLaneIdx(s, t, lane_offset, 0, offset, true, laneTypeMask);
 
                 if (new_lane_index < 0)
                 {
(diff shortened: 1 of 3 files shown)


eknabevcc:
I quickly checked the 2+1 road referred to here:
https://esmini.github.io/#_reference_line_and_center_lane_while_using_laneoffset

Something gets wrong. Now it looks like:

2_plus_one

Could it be that all lane positions (roadId, laneId, offset) should consider lane offset , while road positions (roadId, t) should not?
That's maybe a generic approach that also would solve the sign positioning, since they are based on s, t.

johschmitz:
What you are saying is correct I think. I tried to do something in that direction while changing only the minimum amount of code necessary and without having a good overview at the moment. 

It did work for the example with lane offset that I posted but judging by the picture it could be though that my patch has a sign error for some cases or that we need to apply it differently based on the sign of the offset, the side of the road or so. Let me check that but I might need some help with these nitty gritty details since I am no expert in this code base.

johschmitz:
Okay, I found that it now looks different with odrviewer and the Blender model

image
image

Based on that comparison I think we should be able to find the problem.

johschmitz:
I was able to fix it for the lane surface but still did not figure out the lane markings as they somehow use some OSI code:
image

eknabevcc:
Thanks for the update. Now it's on me to check again. And I'll look into the OSI creation part as well.

eknabevcc:
First shot implementation included in [v2.43.0](https://github.com/esmini/esmini/releases/tag/v2.43.0).

Calculations updated, and graphics looks beautiful (btw: I hope it was OK to add your nice intersection to test case):

Videoclip: https://youtu.be/zGrH35kjyqI

johschmitz:
This looks awesome, I'll have a look at the new release at the beginning of the next year when I get back home.
Sure, no problem to include this example/testcase.

eknabevcc:
Closing since no immediate follow-up issues observed or reported.

Raise a new issue if related issues found in future.

PR #634, opened 2024-11-14
Fix position of wheels

### LINKED ISSUE #633: Position of wheels wrong in OSI
In Release 2.38.2 the position of the wheel was added. Unfortunately, the positions of the wheels seem to be relative to the rear axle. But according to the [OSI specification](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1MovingObject_1_1VehicleAttributes_1_1WheelData.html#abfe9b8868b4a50e8b20476aadd217268) they have to be relative to the center of the bounding box. 

In my test, these are the vehicle attributes of an object, where you can see that the positions of the wheels do not fit to the bbcenter_to_front/rear attributes. x- and z-positions should be the same (or at least similar if deflection is taken into account).

```
vehicle_attributes {
      driver_id {
      }
      bbcenter_to_rear {
        x: -1.1460000276565552
        z: -0.31700000166893005
      }
      bbcenter_to_front {
        x: 1.2540000677108765
        z: -0.31700000166893005
      }
      wheel_data {
        position {
          x: 2.4000000953674316
          y: -0.699999988079071
          z: 0.33000001311302185
        }
        orientation {
          pitch: 4
        }
        friction_coefficient: 1
      }
      wheel_data {
        index: 1
        position {
          x: 2.4000000953674316
          y: 0.699999988079071
          z: 0.33000001311302185
        }
        orientation {
          pitch: 4
        }
        friction_coefficient: 1
      }
      wheel_data {
        axle: 1
        position {
          y: -0.699999988079071
          z: 0.33000001311302185
        }
        orientation {
          pitch: 4
        }
        friction_coefficient: 1
      }
      wheel_data {
        axle: 1
        index: 1
        position {
          y: 0.699999988079071
          z: 0.33000001311302185
        }
        orientation {
          pitch: 4
        }
        friction_coefficient: 1
      }
    }
```

My visualization ignores the wheels with x=0, that's why the rear wheels are displayed in their default position from the 3D model. But the front wheels are visualized according to their position set in OSI.

image

ClemensLinnhoff:
Fixes #633.

Added transformation of the wheel positions to the center of the bounding box.

Now the wheels are where they should be:

image

### DIFF
--- a/EnvironmentSimulator/Unittest/ScenarioEngineDll_test.cpp
+++ b/EnvironmentSimulator/Unittest/ScenarioEngineDll_test.cpp
@@ -333,7 +333,7 @@ TEST(GetOSIRoadLaneTest, lane_no_obj)
     SE_UpdateOSIGroundTruth();
     SE_FlushOSIFile();
     ASSERT_EQ(stat("gt.osi", &fileStatus), 0);
-    EXPECT_EQ(fileStatus.st_size, 83972);  // initial OSI size, including static content
+    EXPECT_EQ(fileStatus.st_size, 84008);  // initial OSI size, including static content
 
     int road_lane_size;
 
@@ -346,13 +346,13 @@ TEST(GetOSIRoadLaneTest, lane_no_obj)
     SE_UpdateOSIGroundTruth();
     SE_FlushOSIFile();
     ASSERT_EQ(stat("gt.osi", &fileStatus), 0);
-    EXPECT_EQ(fileStatus.st_size, 85057);  // slight growth due to only dynamic updates
+    EXPECT_EQ(fileStatus.st_size, 85129);  // slight growth due to only dynamic updates
 
     SE_StepDT(0.001f);  // Step for write another frame to osi file
     SE_UpdateOSIGroundTruth();
     SE_FlushOSIFile();
     ASSERT_EQ(stat("gt.osi", &fileStatus), 0);
-    EXPECT_EQ(fileStatus.st_size, 86143);  // slight growth due to only dynamic updates
+    EXPECT_EQ(fileStatus.st_size, 86251);  // slight growth due to only dynamic updates
 
     SE_DisableOSIFile();
     SE_Close();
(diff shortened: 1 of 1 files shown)


EnvironmentSimulator/Modules/ScenarioEngine/SourceFiles/OSIReporter.cpp
eknabevcc, line 1: Unfortunately I can't edit the file, for some reason (some pull requests allows me to).

I think only the formatting prevents CI going green. Break line after minus sign should fix it, like this:

obj_osi_internal.mobj->mutable_vehicle_attributes()->mutable_wheel_data(i)->mutable_position()->set_x(
    objectState->state_.info.wheel_data[static_cast<unsigned int>(i)].x -
    static_cast<double>(objectState->state_.info.boundingbox.center_.x_));

ClemensLinnhoff: That seems to have worked, thanks!

ClemensLinnhoff:
@eknabevcc could you support with the failed unit tests? I am not entirely sure what they do, to be honest.

ClemensLinnhoff:
> @eknabevcc could you support with the failed unit tests? I am not entirely sure what they do, to be honest.

Okay, I just adjusted those values in the failed unit tests, but I am not sure if this is the correct way to go.

eknabevcc:
Excellent. You did ycorrect adjustment of the test case which checks size of the OSI data, which typically increase as values becomes != 0.

BTW: I like the car model, perhaps inspired by the beauty [p1800](https://www.media.volvocars.com/global/en-gb/heritagemodels/volvo-p1800/227026)  🙂

ClemensLinnhoff:
Thanks for the merge!

The car is the official [ASAM OpenMATERIAL example vehicle](https://github.com/asam-ev/OpenMATERIAL/tree/111-add-example-vehicle-to-demonstrate-vehicle-node-structure/examples/vehicle_example) (currently still on a branch, will be merged shortly). It's under MPL-2.0 license, so feel free to use it :)

PR #643, opened 2024-12-12
Rename EsminiOsiSource to esmini

### LINKED ISSUE #640: Rename EsminiOsiSource.fmu to esmini.fmu
I think it would make sense to rename the OSMP FMU from EsminiOsiSource.fmu to simply esmini.fmu.

The current name was chosen because at the time, the OSMP FMU could only output OSI SensorView messages and did not have any feedback possibility. Now, you can connect traffic participant models, so a closed-loop simulation is also possible. Therefore, it is not "just" an OSI source anymore.

However, would this be considered a break in backwards compatibility and therefore should not be changed?

ClemensLinnhoff:
Rename the OSMP FMU from EsminiOsiSource to esmini.
I also fixed the problem, that lanes are missing in the first time step. This came from calling SE_UpdateOSIGroundTruth(); twice. Once during initialization and then again in the first actual time step.

#Fixes #640


eknabevcc:
Great, thanks!

PR #663, opened 2025-02-03
Update opendrive_16_lane.xsd

Chookees:
Fixed some typos


eknabevcc:
Clearly nice fixes, thanks!

It should here be stated that the OpenX schema files are pulled from the ASAM standards. If this was a more recent version I would recommend to provide your fix to ASAM first. However, OpenDRIVE 1.6 has already bumped once to 1.6.1, so I think there will be no more updates. 

Anyway, let's apply your spelling/typo fixes to the schema file in esmini repo. However, I choose to restore recommended string values, to align with content of the original scheme.

PR #668, opened 2025-02-05
Set refetchStaticGt to true

ClemensLinnhoff:
Enable refetching static ground truth in every time step.


ClemensLinnhoff:
Closing, as this will be handled in a separate branch.

PR #674, opened 2025-02-17
Add pre-commit checks and workflow

doganulus:
This PR introduces `pre-commit` hooks to enforce code quality and consistency before committing changes. It also adds a corresponding CI workflow to check and fix each incoming PR automatically.

Currently, the `esmini` repo does not pass these checks. This can be tested locally by:

```
pre-commit run --all-files
```

after installing `pre-commit` Python library. 

Ref: https://pre-commit.com/

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -6,10 +6,10 @@ on:
       - master
       - dev
       - 'feature/**'
-    tags:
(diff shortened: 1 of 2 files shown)


eknabevcc:
Thanks for the tip to incorporate pre-commit framework. Looks interesting. We'll look into it.

PR #701, opened 2025-05-23
Add wheel rotation rate, radius and number of wheels

### LINKED ISSUE #700: Add rotation rate to OSI wheel data
The rotation rate of the wheels can be used to calculate the a micro-doppler frequency in a sensor simulation.
The data is stored in the [OSI wheel data](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1MovingObject_1_1VehicleAttributes_1_1WheelData.html#aac1cc436ddbf53001ca1a5269b536dc0).

As the orientation of the wheels is set in the OSI ground truth, the information of the rotation rate has to be present in esmini.

ClemensLinnhoff:
I added the rotation rate  and wheel radius to the OSI wheel data. I also added the number of wheels to the vehicle attributes.

Fixes #700

### DIFF
--- a/EnvironmentSimulator/Modules/ScenarioEngine/SourceFiles/Entities.hpp
+++ b/EnvironmentSimulator/Modules/ScenarioEngine/SourceFiles/Entities.hpp
@@ -34,9 +34,9 @@ namespace scenarioengine
         double p = 0.0;  // pitch in global coordinate system
         // double r;                     // roll in global coordinate system
         // double width;                 // median width of the tire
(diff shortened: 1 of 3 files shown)


eknabevcc:
Regarding failing test case. Let me check Monday.
I think it's a simple update of some test cases needed whenever content is added or changed in OSI.
Regarding formatting, we have some updates to be launched very soon making that part simpler using pre-commit framework.

PR #702, opened 2025-05-28
Fix crash when index used is larger than viewer entities_'s size in SteeringSensorSetVisible function

wubinwab:
The SteeringSensorSetVisible function uses object_index to access viewer_->entities_,  but a race condition occurs when a new object is created in the scenario engine loop before the viewer's entity list is updated. This leads to object_index exceeding 
viewer_->entities_.size() and causes a crash.


eknabevcc:
Nice find and fix, thanks!

PR #708, opened 2025-06-13
Add camera interfaces

lrstttl:
Added two new interfaces to esminiLib:

- `SE_GetObjectInCameraFocus()`: returns the ID of the object the viewer camera is currently focused on
- `SE_GetCameraPos()`: returns the viewer camera transform in global coordinates

I still need to make sure if I assigned the correct values from the `osg::Vec3 rot` to the arguments `h`, `p` and `r`.


eknabevcc:
Nice adds, thanks!

I took the liberty of cleaning history slightly, but of course while keeping original author.

Merged on dev and will be included in next release.

PR #727, opened 2025-09-02
doc: fix type indices in documentation

lolainta:
The `enum class ParameterType` defines a `PARAM_TYPE_NONE` entry as the first element.
As a result, all subsequent parameter types are assigned indices starting from 1.

Note: This indexing behavior is currently not covered by unit tests.

Reference: https://github.com/esmini/esmini/blob/3d07b60934cb0824a58cfe2bbafca3b36b0171a2/EnvironmentSimulator/Modules/ScenarioEngine/OSCTypeDefs/OSCParameterDeclarations.hpp#L29-L36

### DIFF
--- a/EnvironmentSimulator/Libraries/esminiLib/esminiLib.hpp
+++ b/EnvironmentSimulator/Libraries/esminiLib/esminiLib.hpp
@@ -672,7 +672,7 @@ extern "C"
     /**
             Get the name of a named parameter
             @param index The index of the parameter, range [0:numberOfParameters-1]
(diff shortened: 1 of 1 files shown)


eknabevcc:
Good find and fix!

In general we avoid stating actual values in the documentation for the very reason that it puts burden. Instead referring to the internal enum for details.  But of course, in the cases where values are documented it should be right. So thanks for fixing.

eknabevcc:
Included in [v2.51.0](https://github.com/esmini/esmini/releases/tag/v2.51.0)

PR #729, opened 2025-09-05
Texture coordinates generation for outline objects

SergeiBorzov:
Side planes have no texture stretching. Top/Bottom planes will have little texture stretching.

image
image


eknabevcc:
Just beautiful and excellent. Good work!

Issue #730, opened 2025-09-17
Ghost trailer has the same name as non-ghost trailer attached to ego


hanze-tan-ai:
When we have an ego with name "ego" and attach a trailer to it. The name of the trailer is "ego+". When we also enable ego ghost, the ghost ego's name is "ego_ghost", but the ghost trailer's name is still "ego+". 

An example of an ego with one trailer attached may look like this:

```
<ScenarioObject name="ego">
  <Vehicle name="vehicle" vehicleCategory="car" model3d="vehicle.fbx" mass="1500">
    <BoundingBox>
      <Center x="1.4" y="0.0" z="0.75" />
      <Dimensions width="2.0" length="5.0" height="1.5" />
    </BoundingBox>
    <Performance maxSpeed="69" maxDeceleration="30" maxAcceleration="200" />
    <Axles>
      <FrontAxle maxSteering="0.5235987756" wheelDiameter="0.8" trackWidth="1.68" positionX="1.422" positionZ="0.37" />
      <RearAxle maxSteering="0.5235987756" wheelDiameter="0.8" trackWidth="1.68" positionX="-1.397" positionZ="0.37" />
    </Axles>
    <TrailerHitch dx="-1.2"/>
     <Trailer>
        <Trailer name="car_trailer">
          <Vehicle name="car_trailer" vehicleCategory="trailer" model3d="../models/car_trailer.osgb">
            <BoundingBox>
              <Center x="0.0" y="0.0" z="0"/>
              <Dimensions width="4" length="15.5" height="4"/>
            </BoundingBox>
            <Performance maxSpeed="69" maxDeceleration="30" maxAcceleration="200"/>
            <Axles>
              <RearAxle maxSteering="0.523598775598" wheelDiameter="0.7" trackWidth="1.68" positionX="-1.397" positionZ="0.37"/>
            </Axles>
            <TrailerCoupler dx="9.25"/>
            <Properties>
              <Property name="model_id" value="11"/>
              <Property name="scaleMode" value="BBToModel"/>
            </Properties>
          </Vehicle>
        </Trailer>
      </Trailer>
    <Properties/>
  </Vehicle>
  <ObjectController>
    <Controller name="ExternalController_ego">
    	  <Properties>
    	     <Property name="esminiController" value="ExternalController" />
    	     <Property name="useGhost" value="true" />
    	     <Property name="headstartTime" value="3" />
    	  </Properties>
    </Controller>
  </ObjectController>
</ScenarioObject>
```

PR #731, opened 2025-09-17
Give ghost trailer correct name

hanze-tan-ai:
This PR clears ghost trailer's name after ghost ego is created, so when ghost trailer is added to Esmini its name can be correctly assigned as `<ghost ego name>+...`.

Closes https://github.com/esmini/esmini/issues/730


eknabevcc:
Great find and fix!
Thanks for contributing.

eknabevcc:
Included in [v2.51.0](https://github.com/esmini/esmini/releases/tag/v2.51.0)

Issue #723, opened 2025-08-01
Disable logging while using libesminiRMLib.so


lukabanic:
I am using the `libesminiRMLib.so` in python and trying to disable its logger. 
Unfortunately, I can not seem to find the way to do it. When setting the argument for the `RM_SetLogFilePath` function to "" as written in the comments of the `esminiRMLib.hpp` file, it still creates the default `esminiRM_log.txt` log file. The older 2.40.4 version works as expected and it seems to me that all the later ones are having an issue. 
Additionaly, the messages are always logged to the `stdout` and I could not find an equivalent function for disabling those logs. I have tried setting the `RM_SetOptionPersistent("--disable_log")` but that did not do the trick.
Is there a way to disable all logging or is this behaviour not predicted when using the `libesminiRMLib.so`?

lukabanic:
Thank you for the bugfix and correcting my workaround attempt 🙂

PR #747, opened 2025-11-20
Feature/osc validation integration

thomassedlmayer:
This is a draft/proposal to include [osc-validation](https://github.com/PMSFIT/osc-validation) in the CI pipeline of ESMini.

Coming from the [European Synergies project](https://synergies-ccam.eu/), we developed an OpenSCENARIO engine validation tool called `osc-validation` (open source). The goal is to detect behavioral differences between scenario engines based on their OSI output but also potential deviations from the OSC standard (for OSC definitions that are unambiguous). Another long-term goal is to identify underspecified or ambiguous parts of the OSC standard.

`osc-validation` is temporarily hosted at https://github.com/PMSFIT/osc-validation. It's soon going to be part of the OpenMSL project.

The only integrated test currently verifies that the vehicle trajectories in the OSI output deviate only marginally from the input data, which ESMini is expected to satisfy.
Though, with the current state of the test case, ESMini would fail because of an integrated ASAM OSI Quality Checker (https://github.com/OpenSimulationInterface/qc-osi-trace) which checks the tool output for compliance with [OSI 3.7.0 rules](https://github.com/OpenSimulationInterface/qc-osi-trace/blob/main/qc_ositrace/checks/osirules/rulesyml/osi_3_7_0.yml) which are based on the `/rules` sections in the proto definitions. The checker result shows that there are some problems with unique assignment of IDs in ESMini's OSI output (different object types have overlapping ID space).

If you are interested in integrating this into ESMini's pipeline, I would ask you to review the CI setup and check out `osc-validation`.

I had to upgrade from Python 3.8 to 3.10 for compatibility reasons. But this should be fine since Python 3.8 already reached its EOL about a year ago. We could even go higher since 3.10 is already close to its EOL timeline.

*Sidenote:*
Also, I noticed that there are some issues with the current pipeline setup when running it in a fork (due to missing tags etc.); see the changes in OSMP_FMU/CMakeLists.txt. This should be removed before ever merging this PR. This is just there so that the pipeline runs properly in my fork.

@jdsika

### DIFF
--- a/.github/actions/run_osc_validation/action.yml
+++ b/.github/actions/run_osc_validation/action.yml
@@ -8,8 +8,9 @@ runs:
       uses: actions/checkout@v4
       with:
         repository: PMSFIT/osc-validation
-        ref: 44c6a2f
+        ref: 44c6a2f4c8363aebf8fcb8e9822fdcca677d51f6
         fetch-depth: 1
+        path: osc-validation
 
     - name: Set up Python venv
       shell: bash
@@ -22,6 +23,7 @@ runs:
         popd
 
     - name: Run osc-validation pytest suite
+      id: run_tests
       shell: bash
       run: |
         pushd osc-validation
--- a/.github/actions/run_osc_validation/action.yml
+++ b/.github/actions/run_osc_validation/action.yml
@@ -1,0 +1,49 @@
+name: "Run osc-validation"
+description: "Install osc-validation, run pytest suite, convert JUnit XML to HTML, and collect artifacts"
+runs:
+  using: "composite"
+  steps:
+
+    - name: Clone osc-validation
+      shell: bash
+      run: |
+        git clone --branch testing --single-branch https://github.com/PMSFIT/osc-validation.git
+
+    - name: Set up Python venv
+      shell: bash
+      run: |
+        pushd osc-validation
+        python3 -m venv .venv
+        source .venv/bin/activate
+        pip install -e .
+        pip install junit2html
+        popd
+
+    - name: Run osc-validation pytest suite
+      shell: bash
+      continue-on-error: true
+      run: |
+        pushd osc-validation
+        source .venv/bin/activate
+
+        SIM_BIN="$(pwd)/../bin/esmini"
+        echo "Using esmini binary: $SIM_BIN"
+
+        mkdir -p osc_validation_results
+
+        pytest validation/scenario/ \
+          --tool ESMini \
+          --toolpath "$SIM_BIN" \
+          -m trajectory \
+          -v \
+          --junitxml=osc_validation_results/junit.xml || true
+
+        cd osc_validation_results
+        junit2html junit.xml junit.html
+        popd
+
+    - name: Upload osc-validation artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: osc-validation
+        path: osc-validation/osc_validation_results
--- a/.github/actions/run_osc_validation/action.yml
+++ b/.github/actions/run_osc_validation/action.yml
@@ -4,10 +4,12 @@ runs:
   using: "composite"
   steps:
 
-    - name: Clone osc-validation
-      shell: bash
-      run: |
-        git clone --branch testing --single-branch https://github.com/PMSFIT/osc-validation.git
+    - name: Checkout osc-validation
+      uses: actions/checkout@v4
+      with:
+        repository: PMSFIT/osc-validation
+        ref: 44c6a2f
+        fetch-depth: 1
 
     - name: Set up Python venv
       shell: bash
--- a/.github/actions/run_osc_validation/action.yml
+++ b/.github/actions/run_osc_validation/action.yml
@@ -21,7 +21,6 @@ runs:
 
     - name: Run osc-validation pytest suite
       shell: bash
-      continue-on-error: true
       run: |
         pushd osc-validation
         source .venv/bin/activate
@@ -31,18 +30,24 @@ runs:
 
         mkdir -p osc_validation_results
 
+        set +e
         pytest validation/scenario/ \
           --tool ESMini \
           --toolpath "$SIM_BIN" \
           -m trajectory \
           -v \
-          --junitxml=osc_validation_results/junit.xml || true
+          --basetemp osc_validation_results \
+          --junitxml osc_validation_results/junit.xml || true
+        pytest_exit_code=$?
 
         cd osc_validation_results
         junit2html junit.xml junit.html
+
         popd
+        exit $pytest_exit_code
 
     - name: Upload osc-validation artifacts
+      if: always()
       uses: actions/upload-artifact@v4
       with:
         name: osc-validation
--- a/.github/actions/run_osc_validation/action.yml
+++ b/.github/actions/run_osc_validation/action.yml
@@ -35,16 +35,16 @@ runs:
           --tool ESMini \
           --toolpath "$SIM_BIN" \
           -m trajectory \
-          -v \
           --basetemp osc_validation_results \
-          --junitxml osc_validation_results/junit.xml || true
-        pytest_exit_code=$?
+          --junitxml osc_validation_results/junit.xml
+        code=$?
+        echo "exitcode=$code" >> $GITHUB_OUTPUT
+        echo "osc-validation finished with exit code $code"
 
         cd osc_validation_results
         junit2html junit.xml junit.html
 
         popd
-        exit $pytest_exit_code
 
     - name: Upload osc-validation artifacts
       if: always()
@@ -52,3 +52,12 @@ runs:
       with:
         name: osc-validation
         path: osc-validation/osc_validation_results
+
+    - name: Fail if tests failed
+      if: always()
+      shell: bash
+      run: |
+        if [ "${{ steps.run_tests.outputs.exitcode }}" != "0" ]; then
+          echo "osc-validation failed"
+          exit 1
+        fi
--- a/.github/actions/run_osc_validation/action.yml
+++ b/.github/actions/run_osc_validation/action.yml
@@ -8,7 +8,7 @@ runs:
       uses: actions/checkout@v4
       with:
         repository: PMSFIT/osc-validation
-        ref: 44c6a2f4c8363aebf8fcb8e9822fdcca677d51f6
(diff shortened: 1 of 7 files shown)


slundel6:
@thomassedlmayer We have an experimental branch which aims to address the issue of overlapping ID's in OSI. Would it be possible for you to verify that branch's functionality against the ASAM OSI Quality Checker? It would be great to get your feedback in case there's still some issues.

The branch in question: [feature/global_osi_ids](https://github.com/esmini/esmini/tree/feature/global_osi_ids)

slundel6:
We added the `host_vehicle_id` (seems like a great thing to have available!) to a patch-release [v2.57.1](https://github.com/esmini/esmini/releases/tag/v2.57.1)

slundel6:
Good catch @ClemensLinnhoff , it was indeed missed. Thanks for the included scenario-files, it made it easy to spot. 

A new commit [81ec9e62](https://github.com/esmini/esmini/commit/81ec9e62da7673b6c6a5cbe02f45231e44e83ae3) has been pushed to dev that hopefully fixes it.

thomassedlmayer:
I updated the osc-validation reference to the newest state including some new test cases to validate start trigger behavior of some exemplary conditions. Good news: esmini passes all of them so far! (see [test (ubuntu-22.04, Release)](https://github.com/thomassedlmayer/esmini/actions/runs/25439726076/job/74627528838#logs) / `Run osc-validation`-step; There you can download the artifacts incl. overall html report.)

I recently noticed that esmini has the `--traj_filter` defaulted to 0.1 which somehow merges close points on a trajectory and thereby slightly manipulates the trajectory. On certain, especially real-world captured trajectories where vehicles are in "almost standstill" this leads to significant increases of the trajectory similarity metric as compared to the input trajectory. For the osc-validation call of esmini I now set this to 0.0. At least for me esmini's output due to the default setting was quite confusing. For the strict follow mode the default setting kind of disregards the OSC spec.

Also, I had to fix some CI issues that occurred out of nowhere because the github actions windows runners seem to report cygwin instead of msys now, causing PYTHON to be unset and smoke_test.py to fail. I don't know if this is just happening on my fork?

@eknabevcc @slundel6 Do you have a timeline for osc-validation to be integrated? Did you have a look if you could agree on the general setup of the github actions for osc-validation? Let me know, if I need to make any changes to comply with your requirements.

Btw, we asked to present the osc-validation project in the OpenSCENARIO CCB, so we get more traction/support also from other scenario engine providers. The discussion will probably take place on May 22.

_Side note: I temporarily disabled the nightly runs because I keep getting notifications due to a missing dev branch on my fork. Maybe these could be set up more fork-friendly in general but I didn't look more into that for now._

eknabevcc:
Thanks @thomassedlmayer for pushing this forward. We'll have a look. 

I can right now confirm I had the same issue with sudden cygwin OSTYPE reported for windows images, but only sometimes. We had to update script (commit d14479429293b7d4edd6c4abf4bede023126a472) to support both "msys" and "cygwin". To my understanding it's just a matter of how the bash environment is launched, it should be no functional difference.

thomassedlmayer:
Hi @eknabevcc,

We released the first version of osc-validation with a few new test cases. It now supports test profiles (`.github/actions/run_osc_validation/test_profile.toml`) which allow you to mark any test cases that are expected to fail for your scenario engine:

```
[[xfail]]
test = "scenario/trajectories/val_interpolation.py::test_timed_polyline_trajectory_interpolation[constant_acceleration_from_initial_speed-*]"
reason = "ESMini does not currently support timed polyline constant-acceleration interpolation."
```

For example, we added a test case validating the new interpolation requirement introduced in OSC 1.4.0 (constant acceleration interpolation), which is expected to fail for esmini right now (see [OSC Issue #723](https://code.asam.net/simulation/openscenario/openscenario-xml/-/issues?show=eyJpaWQiOiI3MjMiLCJmdWxsX3BhdGgiOiJzaW11bGF0aW9uL29wZW5zY2VuYXJpby9vcGVuc2NlbmFyaW8teG1sIiwiaWQiOjk0MjF9)). The test profile lets you ignore test cases you don't want to tackle immediately.

By the way, you also get a comprehensive html report with detailed error logs etc.: https://github.com/thomassedlmayer/esmini/actions/runs/29261950752/artifacts/8283893034

The other xfail-marked tests in the report validate parts of the edge cases defined in section [6.9 of the OSC specification](https://simulation.pages.asam.net/openscenario/openscenario-antora-gen/ASAM_OpenSCENARIO_XML/v1.4.0/06_general_concepts/06_09_trajectories.html#_entity_is_not_at_the_start_of_trajectory_timereference_is_not_set).

slundel6:
Hello @thomassedlmayer ,
We've looked a bit on this PR (finally), and have some basic questions/comments:

- In [test_profile.toml](https://github.com/thomassedlmayer/esmini/blob/bce74298802496e39644d6fb57b6520419c37541/.github/actions/run_osc_validation/test_profile.toml) test [1](https://github.com/thomassedlmayer/esmini/blob/bce74298802496e39644d6fb57b6520419c37541/.github/actions/run_osc_validation/test_profile.toml#L2) and [2](https://github.com/thomassedlmayer/esmini/blob/bce74298802496e39644d6fb57b6520419c37541/.github/actions/run_osc_validation/test_profile.toml#L6) seems appropriately skipped for the time being.
- In [test_profile.toml](https://github.com/thomassedlmayer/esmini/blob/bce74298802496e39644d6fb57b6520419c37541/.github/actions/run_osc_validation/test_profile.toml) test [3](https://github.com/thomassedlmayer/esmini/blob/bce74298802496e39644d6fb57b6520419c37541/.github/actions/run_osc_validation/test_profile.toml#L10) and [4](https://github.com/thomassedlmayer/esmini/blob/bce74298802496e39644d6fb57b6520419c37541/.github/actions/run_osc_validation/test_profile.toml#L14), we can see that esmini jumps to a point on the trajectory (either interpolated between 2 timestamps, or on top of a timestamp depending on the trigger time), which should be the expected behavior. The speed of the vehicle is however not interpolated. Is the speed the root-cause of the failure, or something else?
- To have a smooth start with this, we're considering having this as a separate nightly job (similar to memoryleak and sanitize). This would also make sense if the test-scope grows, and running the checks takes longer and longer time. I've already prepared a branch for this.

Great work so far, feels like a merge is quite close!

thomassedlmayer:
Hi @slundel6!
Thanks for having a look!

Test 3 and 4 fail because esmini's behavior diverges from what section 6.9.3 says in my opinion. I'm not saying that 6.9.3 makes sense but I think the following phrasing to me sounds like the teleport to the corresponding position should happen instantly and not with any delay. 

> [...] the entity behaves as if the trajectory starts at the position interpolated between the most recent time reference and the earliest future time reference. **The entity teleports to that position and continues along the trajectory.** 

Esmini seems to skip that mentioned point and in the subsequent frame jumps to the next trajectory vertex. So from then on the trajectories are in sync again.

I don't want to say that esmini should change any behavior. Instead, our intention was to show that this could be a good process to identify bad definitions in the standard. Similarly, such test cases could be built for new features before publishing them in a new standard version (as we demonstrated for the new acceleration-based interpolation method).

I also created another branch in this repo and removed all the workarounds for the experienced fork CI issues:
https://github.com/esmini/esmini/compare/master...feature/osc-validation-integration

I don't insist on the way I've set up the CI job. Feel free to adapt it as you like!

slundel6:
Thanks for the detailed reply! We're taking the chance to investigate these errors, and I've ran the framework locally (which was smooth following the guide on your repo). 

I don't know this framework very well so I might be doing things wrong, but I find it a bit tricky to find out exactly what went wrong in a failed test. Here's an example (short version) summary of the failed test 3:

>        assert metric_result.sample_count == 5, "Trace should have 5 samples at 1 Hz from 0s to 4s inclusive"
>        assert metric_result.max_time_error <= 0.01, "Trace should have max time error <= 0.01s"
>        assert metric_result.max_xy_error < 0.1, "Trace should have max XY error < 0.1m"
>        AssertionError: Trace should have max XY error < 0.1m
>        assert 15.0 < 0.1
>         +  where 15.0 = ObjectStateMetricResult(reference_object_id=1, tool_object_id=10, sample_count=5, 
>        max_xy_error=15.0, max_planar_speed_...ngth_error=0.0, max_width_error=4.768371586472142e-08, max_height_error=0.0, max_dimension_error=4.768371586472142e-08).max_xy_error
>        osc_validation/validation/scenario/trajectories/val_follow_trajectory_teleport.py:143: AssertionError

I can see that 0.15 is > 0.1 and that this is a distance, and can guess that in this particular case its _probably_ at timestamp 2.0, but is there a way to be certain? E.g. some verbosity level or similar with extra info on actual vs expected values.

thomassedlmayer:
This looks like you most certainly ran the framework correctly! I'm glad it worked fine so far. Let me know, if you're having any issues in the future.

The ObjectState metric is very basic in the way that it just collects a variety of max error values (comparing reference trace and tool-generated trace frame-by-frame) but not the exact moment when each was collected. I agree that it could be a point for improvement to store timestamps as some kind of metadata output of that metric. I think it should be a continuous process to improve these metrics based on the needs we see.

My usual process was to just inspect the generated OSI/mcap traces in a 3D viewer or directly inspect the raw trace content (e.g. object positions). The mcaps are saved to the (probably defaulted) pytest basetemp directory for each validation run. For local validation runs you can use the clickable link in the html report's test case description:
image

As OSI/mcap viewer I usually use Lichtblick with the asam-osi-converter plugin. It lets you also compare two overlayed OSI traces by drag-and-dropping both into a single Lichtblick instance. You can also inspect the raw content using the "Raw Messages" panel and selecting the desired mcap topic name.

My experience was that you usually need to have an in-depth (frame-by-frame) look at the trace files for failed tests anyways in order to understand exactly what's going wrong. So I didn't see the immediate need to implement a more complex metric than necessary. But I see that some more info would definitely help with the intuitiveness.

The metrics are generally something that should be extended as we develop more test cases. I noticed the same for the trajectory similarity metric: Different simulators seem to handle scheduling fairly different which is fine by OSC rules but has some impact on trigger activation. Hence, this requires quite tolerant metrics in the sense that trajectories could be shifted by one or two frames and still be OSC compliant. So we added a lag-aligned trajectory similarity metric which can be used for trigger-activated trajectory comparison.

slundel6:
Thank you for the detailed explanation, very much appreciated.

PR #762, opened 2026-01-06
Copy OSI version from ground truth to sensor view

ClemensLinnhoff:
The OSI version is a mandatory field in all top-level messages but it was missing from the SensorView message created by the OSMP FMU. Now it is copied from the GroundTruth.


eknabevcc:
Nice fix. Thanks.

PR #772, opened 2026-02-10
Fix model ID to 3D model mapping bug

johschmitz:
If model file not explicitly specified set it based on lookup table.


eknabevcc:
Do you have a scenario example pinpointing the issue? I think you have a good patch, but I'd like to clearly understand what case/gap it fixes.

johschmitz:
This is for the case when no 3D models are specified in the vehicle catalog xml. In that case it needs to fall back to use different looking models (different colors) for the different entities based on IDs. There was already some code for that but it was not working properly. Basically just remove the 3d model references from the vehicle catalog to test this.

PR #792, opened 2026-05-04
758 add guard rails and barriers as lane boundaries to osi

### LINKED ISSUE #758: Guard rails and barriers not correctly represented in OSI
In esmini guard rails and barriers are taken from the OpenDRIVE and converted to OSI as stationary objects.
I am currently unsure if this is generally the correct approach to represent barriers in OSI as they can also be represented as a LaneBoundary. I have opened a corresponding issue there: https://github.com/OpenSimulationInterface/open-simulation-interface/issues/894.

But even if using stationary objects with a base_polygon for barriers in OSI would be correct, the current implementation in esmini has 2 problems:

1. The [base_polygon](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1BaseStationary.html#aa1db348acaac2d5a2ba0883903d962cd) in OSI is defined in object coordinates:
  ```
  The polygon is defined in the local object frame: x pointing forward and y to the left. The origin is the center of the bounding box.
  ```
  But in the current implementation the orientation of the object is not taken into account. Let's assume we have a straight section of a barrier on a road that has a certain angle with respect to the world coordinates. Then the stationary object has that angle as an orientation. But the y-coordinates of the base polygon should be close to 0. This is not the case:
  Image
  If I ignore the object orientation, the polygon has the correct shape:
  Image

2. The base_polygon is required to close:
  ```
  As ground truth, the polygon is closed by connecting the last point with the first point.
  ```
  If I do that in a curve, it does not follow the intended shape of the barrier, but it becomes a large object like this:
  Image

ClemensLinnhoff:
Guard rails are currently added to OSI as stationary objects. The problem is, that this way, they cannot have a height profile, so the guard rails will not follow the elevation of the road. A better way is to implement guard rails in OSI as lane boundaries.

There are some open points, that should be discussed:
- Would it be possible to distinguish between a barrier and a guard rail, as these are separate [types](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1LaneBoundary_1_1Classification.html#a9576e0ea6592a1f986bfc9ae2b5a8b47) in OSI.
- Currently the outline of a guard rail always goes to one side and then comes back, so I am taking only half the corner points. Is this generally correct?
- The boundary points do not store the width of the barrier or guard rail. So in this PR, the width is set to a default value of 0.1 m. This would be fine for me, but maybe there is a smarter way to set a width.

Fixes #758


eknabevcc:
I pushed a rebased (on dev) variant of this branch: https://github.com/esmini/esmini/tree/feature/pr792

For future reference, please target dev branch. And avoid merging dev branch. Either just hand over that check to us, or do a rebase on your own (instead of merge). In the end the PR is rebased on dev anyway.

ClemensLinnhoff:
> I pushed a rebased (on dev) variant of this branch: https://github.com/esmini/esmini/tree/feature/pr792
> 
> For future reference, please target dev branch. And avoid merging dev branch. Either just hand over that check to us, or do a rebase on your own (instead of merge). In the end the PR is rebased on dev anyway.

Can that branch be merged?

eknabevcc:
At first glance it seems like a good solution. Hadn't found time to investigate further yet, including synching with internal stakeholders.

PR #807, opened 2026-05-11
Clear unknown_args_ in SE_Options::Reset

### LINKED ISSUE #806: SE_Options::Reset does not clear unknown_args
`SE_Options::Reset()` clears `optionOrder_`, per-option `set_/arg_value_`, and `originalArgs_,` but does not clear unknown_args_. Stale entries from a previous `ParseArgs` survive `Reset()`, so `HasUnknownArgs()` keeps returning true on subsequent runs even when the new argument set is clean.

Affects any process that initializes esmini more than once in the same address space — Python bindings, embedded library users, repeated `SE_Init / SE_Close` cycles. 

Reproduction:

1. `SE_InitWithArgs(... "--xyzzy" ...) → HasUnknownArgs() → true` (expected).
2. `SE_Close().`
3. `SE_InitWithArgs(... only valid args ...) → HasUnknownArgs() still true`; "Unrecognized arguments:" prints stale "--xyzzy"

mirkobulaja-avl:
### Summary
`SE_Options::Reset()` did not clear unknown_args_. Stale entries from a previous ParseArgs cycle survived the reset, causing HasUnknownArgs() to return true and PrintUnknownArgs() to print stale arguments on subsequent runs even when the new args are clean.

### Fix
One-line addition in `SE_Options::Reset()`

### Impact
Affects any process that re-initializes esmini in the same address space (Python bindings, embedded library users, `SE_Init → SE_Close → SE_Init`). `SE_Close` does not call` Options::Reset()`; it is invoked at the start of the next `SE_Init` (playerbase.cpp:126), so the stale state is directly observable across re-initialization.

Closes #806

### DIFF
--- a/EnvironmentSimulator/Modules/CommonMini/CommonMini.cpp
+++ b/EnvironmentSimulator/Modules/CommonMini/CommonMini.cpp
@@ -2789,6 +2789,7 @@ void SE_Options::Reset()
         }
     }
     originalArgs_.clear();
+    unknown_args_.clear();
 }
 
 int SE_WritePPM(const char* filename, int width, int height, const unsigned char* data, int pixelSize, int pixelFormat, bool upsidedown)


eknabevcc:
Thanks for good find and fix!

PR #809, opened 2026-05-11
Guard against null Route::GetWaypoint() in route-relative motion

mirkobulaja-avl:
## Summary

`Route::GetWaypoint()` returns `nullptr` when the route has no minimal waypoints or when `waypoint_idx_` is `IDX_UNDEFINED` (e.g. after the entity is reported off-route, or when `AcquirePositionAction`'s target waypoint cannot be resolved). Several call sites dereferenced the result unconditionally and could segfault mid-simulation.

## Changes

- `Position::SetRouteLanePosition` / `SetRouteRoadPosition` — log error and fall back to default direction when waypoint is null (only triggered with `--align_routepositions`).
- `Route::MovePathDS` — return `ERROR_GENERIC` instead of dereferencing.
- `Route::SetPathS` — handle null waypoint in the "moved out of route" log, and return `ERROR_GENERIC` instead of touching waypoint 0 when `minimal_waypoints_` is empty.
- `AssignRouteAction::Start` / `AcquirePositionAction::Start` — skip route assignment with `LOG_WARN` when the constructed route is invalid, rather than installing it on the entity (which would defer the crash to the next route-motion call).

## Notes / caveats

- The vulnerable code paths are present in the source, but I was not able to deterministically reproduce a segfault using only the bundled scenarios. The crash was originally observed in a customer scenario with a different road network where `waypoint_idx_` becomes `IDX_UNDEFINED` while subsequent route-relative motion triggers the segfault.

### DIFF
--- a/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
+++ b/EnvironmentSimulator/Modules/RoadManager/RoadManager.cpp
@@ -12127,7 +12127,15 @@ int Position::SetRouteLanePosition(Route* route, double path_s, int lane_id, dou
     int dir = 1;
     if (SE_Env::Inst().GetOptions().GetOptionSet("align_routepositions"))
     {
-        dir = route->GetWaypoint()->GetRouteWaypointDir();
+        const Position* wp = route->GetWaypoint();
+        if (wp != nullptr)
+        {
+            dir = wp->GetRouteWaypointDir();
+        }
+        else
+        {
+            LOG_ERROR("SetRouteLanePosition: failed to get current waypoint, using default direction");
+        }
     }
 
     SetLanePos(route->GetTrackId(), SIGN(dir) * lane_id, route->GetTrackS(), SIGN(dir) * lane_offset);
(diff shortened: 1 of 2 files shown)


eknabevcc:
Very good find. And thanks for the fix(es)!

I saw first a few test cases failing by the initial commit. I suggest a slight modification of the changes for private actions (commit f39e50dc3c2798a1063e1d5ac58b5f4bc26f90b9) .

It's a bit complicated. To support the [FollowRouteController](https://esmini.github.io/controllers.html#_followroute) (which can find paths between disconnected waypoints by injecting lane changes) two sets of waypoints are handled in parallel. One for default controller and one for custom controllers. Also the original set, as specified in the scenario, are stored for reference. See [here in RoadManager.hpp](https://github.com/esmini/esmini/blob/a8e18fce72e283e9b8b5ae2183d8f8faad34aecb/EnvironmentSimulator/Modules/RoadManager/RoadManager.hpp#L4985-L4987).

So, basically by allowing routes to be assigned to entities, even if they are "invalid" (no connected path found for default controller) still custom controllers can make use of the defined route.

I haven't investigated fully, but I think by the proposed additional commit we have a perfect baby step forward that at least solves the most immediate bug causing crash via nullptr.

What do you think? I you agree I'll merge the PR.

mirkobulaja-avl:
Changes look good to me, verified manually in our setup, and also built and run the tests this time, no difference to 'dev' branch.

PR #821, opened 2026-07-02
Add Object::TimeToCollision and SE_GetTimeToCollision API

kosuke55:
Hi, thanks for the great project!

Add time-to-collision computation as `Object::TimeToCollision` and expose it via `SE_GetTimeToCollision` in esminiLib.
`TrigByTimeToCollision::CheckCondition` delegates the object-to-object path to the new method (no behavior change).

PR #824, opened 2026-07-19
Expose TTC computation to esminiJS API

kosuke55:
Follow-up to #821, which added TTC to the esminiLib C API.

This adds the same `Object::TimeToCollision` computation to the esminiJS (WASM)
API, so it can be called from the browser build via `get_ttc`.

### DIFF
--- a/EnvironmentSimulator/Libraries/esminiJS/esminijs.cpp
+++ b/EnvironmentSimulator/Libraries/esminiJS/esminijs.cpp
@@ -675,6 +675,32 @@ namespace esmini
         return copy_state_from_object(object, scenario_engine_->getSimulationTime());
     }
 
+    double OpenScenario::get_ttc(int object_a_id, int object_b_id, bool free_space, int cs, int dist_type) const
+    {
+        if (scenario_engine_ == nullptr)
+        {
+            return -1.0;
+        }
+
+        scenarioengine::Object* obj_a = scenario_engine_->entities_.GetObjectById(object_a_id);
+        scenarioengine::Object* obj_b = scenario_engine_->entities_.GetObjectById(object_b_id);
+        if (obj_a == nullptr || obj_b == nullptr)
+        {
+            return -1.0;
+        }
+
+        double ttc = -1.0;
+        if (obj_a->TimeToCollision(obj_b,
+                                   static_cast<roadmanager::CoordinateSystem>(cs),
+                                   static_cast<roadmanager::RelativeDistanceType>(dist_type),
+                                   free_space,
+                                   ttc) != 0)
+        {
+            return -1.0;
+        }
+        return ttc;
+    }
+
     ScenarioRoadGeometry OpenScenario::collect_road_geometry() const
     {
         ScenarioRoadGeometry geometry;
--- a/EnvironmentSimulator/Libraries/esminiJS/esminijs.cpp
+++ b/EnvironmentSimulator/Libraries/esminiJS/esminijs.cpp
@@ -675,7 +675,7 @@ namespace esmini
         return copy_state_from_object(object, scenario_engine_->getSimulationTime());
     }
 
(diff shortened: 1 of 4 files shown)


kosuke55:
@johschmitz 
Since this touches the esminiJS bindings you built in #814, would you be able to take a look when you have a moment? It just exposes the TTC computation (added in #821) through get_ttc. Thanks!

EnvironmentSimulator/Libraries/esminiJS/esminijs.cpp
```diff
@@ -675,6 +675,32 @@ namespace esmini
         return copy_state_from_object(object, scenario_engine_->getSimulationTime());
     }
 
+    double OpenScenario::get_ttc(int object_a_id, int object_b_id, bool free_space, int cs, int dist_type) const
```
johschmitz: For consistency with the rest of the codebase I suggest time_to_collision() since
I understand this is a wrapper for TimeToCollision()

EnvironmentSimulator/Libraries/esminiJS/esminijs.cpp
```diff
@@ -675,6 +675,32 @@ namespace esmini
...
+    {
+        if (scenario_engine_ == nullptr)
+        {
+            return -1.0;
+        }
+
+        scenarioengine::Object* obj_a = scenario_engine_->entities_.GetObjectById(object_a_id);
+        scenarioengine::Object* obj_b = scenario_engine_->entities_.GetObjectById(object_b_id);
+        if (obj_a == nullptr || obj_b == nullptr)
```
johschmitz: Lots of return -1 extra checks. I think not all of these checks are needed, some of it is handled inside the wrapped method and maybe some conditions can never occur. IMHO add necessary checks to avoid crashes but other than that try to make the wrapper as thin as possible and if necessary add additional checks to the wrapped method so that WASM ans non-WASM builds behave as similar as possible.

EnvironmentSimulator/Libraries/esminiJS/esminijs.cpp
kosuke55: thanks, that is better.
Renamed to `get_time_to_collision()` keeping the get_ prefix to match the other esminiJS getters.
a3dfb0d6dbc6de37a62474978c7434b760290019

EnvironmentSimulator/Libraries/esminiJS/esminijs.cpp
kosuke55: made it thinner and kept only the null checks needed to avoid a crash.
dc9b1a48a24694f9476a937e9ea5b0fd3ad54572

EnvironmentSimulator/Libraries/esminiJS/esminijs.cpp
johschmitz: Not 100% sure how consistent this is being handled in this project (maybe we need to cross check) but according to my most common understanding getters just access a value of a single object while this method is performing calculations across two object's data. Hence, IMHO for this particular method the get_ is somewhat misleading/confusing (unless some other code constantly calculates all TTCs towards all other objects and then stores a huge array of these values alongside every object and that is what we are accessing here).

kosuke55: Thanks, agreed. Not a pure getter, so dropped get_. 
a7caf64d2cf069ccf0ea984232646eb6c07e0b7d

kosuke55:
@johschmitz @eknabevcc 
No rush, but please let me know if there's anything else I need to address on my side :bow:

johschmitz (approved):

johschmitz:
LGTM

eknabevcc:
LGTM too.
Thanks @kosuke55 for the contribution and thanks @johschmitz for the active review!

Issue #357, opened 2022-11-25
WebAssembly build Support


mljack:
Hello,
Does anyone try to build esmini with emscripten? It would be amazing to run the scenario simulation toolchains in a browser!

Maybe it's eaiser to start with esminiLib and esminiRMLib which have less dependency libraries. And just found OSG itself already has an [example](https://github.com/openscenegraph/OpenSceneGraph/tree/master/examples/osgemscripten) for emscripten.

Cheer,
mljack

eknabevcc:
I'm noob when it comes to Java and emscripten. But actually there is a very interesting pull request [PR 356](https://github.com/esmini/esmini/pull/356) that I think address exactly this goal, to run in browser - which would be very cool :)

Unfortunately I have not yet succeeded in building the code in the pull request. I'm waiting for some guidance before trying again. 

Hopefully that pull request can be a start of a web-based esmini.

mljack:
Great! Thanks for your kindly informative noticing! I'll have a try later this week.

zn845639326:
I did similar work in the first half of last year. The simulation scenario is successfully run in the browser and the log is printed in the console. I didn't change the esmini core code. Only some of the CMakeLists were modified to ensure that the compilation passed. The implementation was a little hard, and the running scenario was a little limited, so I didn't publish it. Below is a screenshot of cut-in scenario running:
![image](https://github.com/esmini/esmini/assets/23354015/c7424bce-2ced-4516-b913-aaed2de38fe3)

eknabevcc:
Cool, looks like it works great!

The [pull request #356](https://github.com/esmini/esmini/pull/356) has been merged. You can find some info here: 
https://github.com/esmini/esmini/tree/master/EnvironmentSimulator/Libraries/esminiJS

So far it's not supporting 3D graphics in the browser, just executing the scenario headless. And I think there also are some limitations, e.g. how to provide the scenario file.

And yes, it would be fantastic to run esmini in a browser WITH the (OSG based) viewer as well. So, if anyone fins a solution let us know 👍

PR #825, opened 2026-07-24
Fix standalone esminiJS WASM build: generate CommonMini/version.cpp

jdsika:
## Problem

Building the standalone **esminiJS** Emscripten/WASM target
(`EnvironmentSimulator/Libraries/esminiJS`, per its `build.sh` / `CMakeLists.txt`)
compiles all objects but then **fails to link**:

```
undefined symbol: ESMINI_GIT_REV
undefined symbol: ESMINI_GIT_TAG
undefined symbol: ESMINI_GIT_BRANCH
undefined symbol: ESMINI_BUILD_VERSION
```

Those symbols are referenced by `CommonMini.cpp` and *defined* in a generated
`version.cpp` (produced from `version.cpp.in` by `CommonMini/version.cmake`). The
full esmini build generates it via `CommonMini/CMakeLists.txt`, but the standalone
`esminiJS` target never triggers that generation, so `version.cpp` is missing from
the link.

## Fix

Invoke `version.cmake` from the `esminiJS` `CMakeLists.txt` exactly the way
`CommonMini/CMakeLists.txt` already does
(`cmake -DESMINI_BUILD_VERSION=... -P version.cmake`). The existing
`COMMONMINI_SOURCES` glob then picks up the generated `CommonMini/version.cpp`, so
no source-list change is needed. 20 lines, one file.

Under a shallow checkout `git describe` yields nothing and the values resolve to
`"N/A"`, which is harmless.

## Verification

With this change the `esminiJS` target builds and links standalone via
`emcmake cmake -G Ninja -DCMAKE_BUILD_TYPE=Release` + `SINGLE_FILE=1`, producing a
working `esmini.js`. I ran it in headless Chromium: `get_road_geometry()` returns
lane surfaces / lane centers and `step_frame()` advances entities, with `reset()`
restarting the run (endless-loop playback).

Refs #357.

## Note

Happy to follow up on distribution (an npm package / a prebuilt WASM release asset)
if that's useful — there is currently no published artifact, which is the main thing
keeping downstream projects from consuming esminiJS directly.

### DIFF
--- a/EnvironmentSimulator/Libraries/esminiJS/CMakeLists.txt
+++ b/EnvironmentSimulator/Libraries/esminiJS/CMakeLists.txt
@@ -17,6 +17,24 @@ set(CMAKE_CXX_STANDARD
 
 project(${TARGET})
 
+# Generate CommonMini/version.cpp for the standalone esminiJS build.
+#
+# CommonMini.cpp references the symbols ESMINI_GIT_REV, ESMINI_GIT_TAG, ESMINI_GIT_BRANCH and ESMINI_BUILD_VERSION, which are *defined* in a generated
+# version.cpp (produced from version.cpp.in by CommonMini/version.cmake). The full esmini build generates it via CommonMini/CMakeLists.txt; this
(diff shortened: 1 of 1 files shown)


eknabevcc:
Nice fix, thanks!

PR #835, opened 2026-08-21
Fix direct junction validation logging

mirkobulaja-avl:
Fixes a fmt::format_error (argument not found) when validating a direct junction whose linked road points back to a different junction.

The diagnostic contained two format placeholders but supplied only one argument. This triggered an exception when using OpenDRIVE 1.9 example Ex_Slip_Lane.xodr (example contains an error).


eknabevcc:
Nice find and fix!

PR #837, opened 2026-09-02
Issue 836

### LINKED ISSUE #836: Configurable UDP port and packet data size for OSI output
### Issue

Currently, when sending OSI GroundTruth messages over UDP using `--osi_receiver_ip`, the destination port is hardcoded to `48198`, and the maximum data payload per packet before fragmentation is fixed to `8192` bytes.
These parameters may be exposed and made configurable.

### Proposed Changes
 
1. Add CLI Optional arguments for port and data size (which is not packet size)
2. Add `SE_OpenCustomOSISocket(const char *ipaddr, unsigned int port, unsigned int data_max_size)` in `esminiLib`.
3. Add automated test coverage to verify UDP packet transmission, packet sizing, and warning messages.

### Design Trade-offs & Discussion Points

- The **simplest** solution to support UDP datagram payloads up to the maximum theoretical size is to set the internal `osi_udp_buf` static buffer from 8 KB to ~64 KB. I think it is a reasonable tradeoff.
- The `osi_udp_buf` struct (containing `counter`, `datasize`, and `data[]`) is currently scoped locally in `OSIReporter.cpp`. If we avoid breaking encapsulation by moving internal implementation details into `OSIReporter.hpp`, the min/max limits must be defined on data size rather then packet size. Header size calculation must be handled somewhat manually when the struct changes schema. I do not foresee different approaches for now...
- `osi_receiver.cpp` too should be updated to support the larger 64 KB buffer size. This is a simple edit.
- I would like to collect community feedback on how to make `osireceiver` configurable for receive port:
  - Command-line arguments (e.g., `--port <port>`)?
  - Environment variables (e.g., `ESMINI_OSI_PORT` / `OSI_RECEIVER_PORT`)? (simpler, but diverges from current `esmini` binary implementation)
- Making the UDP port, buffer sizing, and socket opening configurable is also a preparatory step towards introducing **UDP Multicast** support in future updates, enabling multiple listeners to subscribe to the same OSI stream without duplicating simulation sender instances.

MatteoRagni:
### Description

This PR adds support for configuring both the UDP port and the maximum UDP data payload size for OSI output transmission in esmini, and introduces a dedicated test suite to validate these options.

Closes #836 

### Changes

- **CLI Options**: Added `--osi_receiver_port <port>` (valid range: 1024..65536, default: 48198) and `--osi_receiver_udp_data_size <size>` (valid range: 1464..65499, default: 8192). Implementation uses same defaults as before.
- **C API**: Exposed `SE_OpenCustomOSISocket(const char *ipaddr, unsigned int port, unsigned int data_max_size)` in `esminiLib` while preserving backwards compatibility with `SE_OpenOSISocket`. **Is the name for the public api ok?**.
- Added bounds checks and warning logs in `OSIReporter` when requested port or data size values fall outside allowable ranges. I hope it is clear enough...
- Updated `osi_receiver.cpp` buffer size to handle packets up to 64 KB.
- **C#**: updated `ESMiniWrapper.cs`

**Am I missing something?**

### Testing

I added new test `test/osi_udp_test.py` (and registered it in `scripts/run_tests.sh`) covering:

1. Verifies UDP transmission on custom port `127.0.0.1:5000`.
2. Verifies that received packets respect the requested data size threshold (e.g., 1500 bytes) and valid 8-byte header layout (`counter`, `datasize`).
3. Verifies default behavior on port 48198 and standard 8192-byte limit when no flags are supplied.
4. Verifies that out-of-range ports trigger appropriate warning logs and fallback to default.
5. Verifies that below-minimum and above-maximum data size inputs trigger clamping warning logs.

I'll leave aside multicast implementation. It mainly impacts `osi_receiver.cpp`, and it needs the implementation of a "configuration" for it. See #836 for more info.

### Additional Notes

- On OSX, maximum datagram size is defined by sysctl `"net.inet.udp.maxdgram"`. This information is collected at runtime to clamp the maximum data size. Tests also consider this.
- There was a TODO note for @eknabe [here](https://github.com/esmini/esmini/blob/dev/EnvironmentSimulator/Modules/ScenarioEngine/SourceFiles/OSIReporter.cpp#L421) regarding a possible overflow. I handled it with input sanitization. Data size cannot be larger than theoretical UDP datagram size (64k and dimes) so it can never overflow.

### DIFF
--- a/EnvironmentSimulator/Applications/replayer/osi_receiver.cpp
+++ b/EnvironmentSimulator/Applications/replayer/osi_receiver.cpp
@@ -22,7 +22,7 @@
 
 #ifndef _WINDOWS
 #include <unistd.h>
-#define Sleep(x) usleep((x) * 1000)
+#define Sleep(x) usleep((x)*1000)
 #endif
 
 static bool quit;
--- a/EnvironmentSimulator/Applications/replayer/osi_receiver.cpp
+++ b/EnvironmentSimulator/Applications/replayer/osi_receiver.cpp
@@ -22,7 +22,7 @@
 
 #ifndef _WINDOWS
 #include <unistd.h>
-#define Sleep(x) usleep((x)*1000)
+#define Sleep(x) usleep((x) * 1000)
 #endif
 
 static bool quit;
--- a/EnvironmentSimulator/Applications/replayer/osi_receiver.cpp
+++ b/EnvironmentSimulator/Applications/replayer/osi_receiver.cpp
@@ -45,7 +45,7 @@ typedef int SE_SOCKET;
 #define OSI_OUT_PORT          48198
 #define ES_SERV_TIMEOUT       500
 #define MAX_MSG_SIZE          1024000
-#define OSI_MAX_UDP_DATA_SIZE 8200
+#define OSI_MAX_UDP_DATA_SIZE 65536
 
 void CloseGracefully(SE_SOCKET socket)
 {
--- a/EnvironmentSimulator/Libraries/esminiLib/ESMiniWrapper.cs
+++ b/EnvironmentSimulator/Libraries/esminiLib/ESMiniWrapper.cs
@@ -1847,6 +1847,13 @@ namespace ESMini
         public static extern int SE_OpenOSISocket(string ipaddr);
 
         /// <summary>
+        /// Send OSI packages over UDP to specified IP address, port and with a threshold for
+        /// maximum size for each OSI data package. Value bigger than data_max_size will be split into multiple packages.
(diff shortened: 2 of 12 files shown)


MatteoRagni:
Ok I think I need some help on (windows-2022, Debug). It is failing on something that was not modified by this pull request (road-model.vcxprj)... @eknabe do you have suggestions?

MatteoRagni:
> Ok I think I need some help on (windows-2022, Debug). It is failing on something that was not modified by this pull request (road-model.vcxprj)... @eknabe do you have suggestions?

Found...

